PR #12, opened 2020-02-08
Fix go.mod

at-wat:
Remove local replace.


lherman-cs:
Oops, totally forgot removing that. Thanks

PR #13, opened 2020-02-08
Implement vpx (VP8 and VP9) encoder

at-wat:
Requires libvpx.

### DIFF
--- a/pkg/codec/vpx/vpx.go
+++ b/pkg/codec/vpx/vpx.go
@@ -1,0 +1,182 @@
+// Package vpx implements VP8 and VP9 encoder.
+// This package requires libvpx headers and libraries to be built.
+package vpx
+
+// #cgo pkg-config: vpx
+// #include <stdlib.h>
+// #include <vpx/vpx_encoder.h>
+// #include <vpx/vpx_image.h>
+// #include <vpx/vp8cx.h>
+//
+// // C function pointers
+// vpx_codec_iface_t *ifaceVP8() {
+//   return vpx_codec_vp8_cx();
+// }
+// vpx_codec_iface_t *ifaceVP9() {
+//   return vpx_codec_vp9_cx();
+// }
+//
+// // C union helpers
+// void *pktBuf(vpx_codec_cx_pkt_t *pkt) {
+//   return pkt->data.frame.buf;
+// }
+// int pktSz(vpx_codec_cx_pkt_t *pkt) {
+//   return pkt->data.frame.sz;
+// }
+//
+// // Alloc helpers
+// vpx_codec_ctx_t *newCtx() {
+//   return malloc(sizeof(vpx_codec_ctx_t));
+// }
+// vpx_image_t *newImage() {
+//   return malloc(sizeof(vpx_image_t));
+// }
+import "C"
+
+import (
+	"errors"
+	"fmt"
+	"image"
+	"io"
+	"unsafe"
+
+	"github.com/pion/mediadevices/pkg/codec"
+	mio "github.com/pion/mediadevices/pkg/io"
+	"github.com/pion/mediadevices/pkg/io/video"
+	"github.com/pion/mediadevices/pkg/prop"
+
+	"github.com/pion/webrtc/v2"
+)
+
+type encoder struct {
+	codec            *C.vpx_codec_ctx_t
+	raw              *C.vpx_image_t
+	r                video.Reader
+	frameIndex       int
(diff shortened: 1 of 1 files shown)


pkg/codec/vpx/vpx.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+		return n, err
+	}
+
+	img, err := e.r.Read()
+	yuvImg := img.(*image.YCbCr)
+
+	e.raw.planes[0] = (*C.uchar)(&yuvImg.Y[0])
+	e.raw.planes[1] = (*C.uchar)(&yuvImg.Cb[1])
+	e.raw.planes[2] = (*C.uchar)(&yuvImg.Cr[2])
```
lherman-cs: I think they should be `&yuvImg.Cb[0]` and `&yuvImg.Cr[0]`

lherman-cs:
Btw, the color seems to be off. But, I really don't mind it. This is definitely a good start! And, I'm sorry for the mess...

Screenshot from 2020-02-08 10-20-10

pkg/codec/vpx/vpx.go
```diff
@@ -0,0 +1,182 @@
...
+			e.raw.fmt = C.VPX_IMG_FMT_I444
+		case image.YCbCrSubsampleRatio422:
+			e.raw.fmt = C.VPX_IMG_FMT_I422
+		case image.YCbCrSubsampleRatio420:
+			e.raw.fmt = C.VPX_IMG_FMT_I420
+		case image.YCbCrSubsampleRatio440:
+			e.raw.fmt = C.VPX_IMG_FMT_I440
+		default:
+			return 0, fmt.Errorf("unsupported pixel format: %s", yuvImg.SubsampleRatio)
+		}
```
lherman-cs: Is this probably the reason why the color is off? Just curious.

at-wat: VP8 looks only supporting I420 and YV12. (VP9 supports I422, I444 and I440 on profile1, but pion/rtp doesn't have payloader yet.)
On my environment, image from camera is I422.

I'll add convert to make it work.

pkg/codec/vpx/vpx.go
at-wat: oops, fixed.

at-wat:
@lherman-cs fixed the pointer and the color problem. Please take another look.

pkg/codec/vpx/vpx.go
```diff
@@ -0,0 +1,201 @@
...
+	e.raw.planes[0] = (*C.uchar)(&yuvImg.Y[0])
+	e.raw.planes[1] = (*C.uchar)(&yuvImg.Cb[0])
+	e.raw.planes[2] = (*C.uchar)(&yuvImg.Cr[0])
+	e.raw.stride[0] = C.int(yuvImg.YStride)
+	e.raw.stride[1] = C.int(yuvImg.CStride)
+	e.raw.stride[2] = C.int(yuvImg.CStride)
+	h := yuvImg.Rect.Max.Y - yuvImg.Rect.Min.Y
+
+	// Covert pixel format to I420
+	// TODO: maybe better to move it to a proper module
```
lherman-cs: Definitely agree. It looks like we can put this under `io/video` in the future. The pattern will look like a middleware. For example,

```
func ToI420(r video.Reader) video.Reader
```

pkg/codec/vpx/vpx.go
lherman-cs: Ah, I see.

pkg/codec/vpx/vpx.go
```diff
@@ -0,0 +1,201 @@
...
+		for i := 0; i < h/2; i++ {
+			addrSrc := i * 2 * yuvImg.CStride
+			addrDst := i * yuvImg.CStride
+			for j := 0; j < yuvImg.CStride; j++ {
+				cb := uint16(yuvImg.Cb[addrSrc+j]) + uint16(yuvImg.Cb[addrSrc+yuvImg.CStride+j])
+				cr := uint16(yuvImg.Cr[addrSrc+j]) + uint16(yuvImg.Cr[addrSrc+yuvImg.CStride+j])
+				yuvImg.Cb[addrDst+j] = uint8(cb / 2)
+				yuvImg.Cr[addrDst+j] = uint8(cr / 2)
+			}
+		}
```
lherman-cs: Wow, I didn't expect that would implement them now. Thanks a lot ^^

PR #15, opened 2020-02-09
Add Firefox and VP8 support to the simple example

at-wat:
Now, it should work both on Firefox and Chrome.

### DIFF
--- a/examples/simple/main.go
+++ b/examples/simple/main.go
@@ -7,7 +7,13 @@ import (
 	"github.com/pion/mediadevices/examples/internal/signal"
 	_ "github.com/pion/mediadevices/pkg/codec/openh264" // This is required to register h264 video encoder
 	_ "github.com/pion/mediadevices/pkg/codec/opus"     // This is required to register opus audio encoder
+	_ "github.com/pion/mediadevices/pkg/codec/vpx"
+	"github.com/pion/mediadevices/pkg/frame"
 	"github.com/pion/webrtc/v2"
+)
+
+const (
+	videoCodecName = webrtc.VP8
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


examples/simple/main.go
```diff
@@ -39,9 +54,10 @@ func main() {
 			c.Enabled = true
 		},
 		Video: func(c *mediadevices.MediaTrackConstraints) {
-			c.Codec = webrtc.H264
+			c.Codec = videoCodecName
+			c.FrameFormat = frame.FormatI420 // most of the encoder accepts I420
 			c.Enabled = true
-			c.Width = 800
+			c.Width = 640
```
at-wat: I think most of the camera supports 640x480. Mine doesn't support 800x480...

examples/simple/main.go
```diff
@@ -7,9 +7,15 @@ import (
 	"github.com/pion/mediadevices/examples/internal/signal"
 	_ "github.com/pion/mediadevices/pkg/codec/openh264" // This is required to register h264 video encoder
 	_ "github.com/pion/mediadevices/pkg/codec/opus"     // This is required to register opus audio encoder
+	_ "github.com/pion/mediadevices/pkg/codec/vpx"
+	"github.com/pion/mediadevices/pkg/frame"
 	"github.com/pion/webrtc/v2"
 )
 
+const (
+	videoCodecName = webrtc.VP8
```
at-wat: Revert it to the original one, H264.

at-wat: Oh, firefox doesn't support H264.

lherman-cs: Oh really. I didn't know that. I was wondering why the example didn't work for H264.... Thanks

examples/simple/main.go
lherman-cs: I see. I guess my camera is the weird one, lol

examples/simple/main.go
```diff
@@ -56,9 +72,10 @@ func main() {
 		}
 	}
 
-	// Wait for the offer to be pasted
-	offer := webrtc.SessionDescription{}
-	signal.Decode(signal.MustReadStdin(), &offer)
+	// Tweak transceiver direction to work with Firefox
+	for _, t := range peerConnection.GetTransceivers() {
+		t.Direction = webrtc.RTPTransceiverDirectionSendonly
```
lherman-cs: Ah, this is new to me too. I didn't know that Firefox requires the directions. Thanks

examples/simple/main.go
```diff
@@ -19,8 +25,17 @@ func main() {
...
 	}
 
+	// Wait for the offer to be pasted
+	offer := webrtc.SessionDescription{}
+	signal.Decode(signal.MustReadStdin(), &offer)
+
 	// Create a new RTCPeerConnection
-	peerConnection, err := webrtc.NewPeerConnection(config)
+	mediaEngine := webrtc.MediaEngine{}
+	if err := mediaEngine.PopulateFromSDP(offer); err != nil {
```
lherman-cs: This makes sense.

at-wat:
> The audio seems to be not working on Firefox.

Oh, really. On my linux environment, audio also works fine.

PR #17, opened 2020-02-09
Add ToI420 converter

lherman-cs:
Moved I420 converter from vpx.go, converted `video.Reader` in openh264.go, and created ToI420 to convert a `video.Reader` to a new `video.Reader` that will emit images in I420 format.


lherman-cs:
> LGTM!
> 
> You might also like to add it to h264:
> https://github.com/pion/mediadevices/blob/e4da8fae14ba6c1e0e7d94fd339f77898fd4bb32/pkg/codec/openh264/openh264.go#L72-L81

Good catch. I've updated the PR to include it.

PR #28, opened 2020-02-10
Add OnEnded handler

### LINKED ISSUE #18: Add device/codec error handler
In Web API, `MediaStreamTrack.ended` event is fired and `MediaStreamTrack.onended` handler is called on such errors.
https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/onended
> This event occurs when the track will no longer provide data to the stream for any reason, including the end of the media input being reached, the user revoking needed permissions, the source device being removed, or the remote peer ending a connection.

https://github.com/pion/mediadevices/blob/e4da8fae14ba6c1e0e7d94fd339f77898fd4bb32/track.go#L115-L124
https://github.com/pion/mediadevices/blob/e4da8fae14ba6c1e0e7d94fd339f77898fd4bb32/track.go#L181-L185

at-wat:
In Web API, `MediaStreamTrack.ended` event is fired and
`MediaStreamTrack.onended` handler is called on device error.
This commit adds same way to handle errors.

### Reference issue
Fixes #18

### DIFF
--- a/examples/simple/main.go
+++ b/examples/simple/main.go
@@ -66,7 +66,12 @@ func main() {
 	}
 
 	for _, tracker := range s.GetTracks() {
-		_, err = peerConnection.AddTrack(tracker.Track())
+		t := tracker.Track()
+		tracker.OnEnded(func(err error) {
+			fmt.Printf("Track (ID: %s, Label: %s) ended with error: %v\n",
+				t.ID(), t.Label(), err)
+		})
+		_, err = peerConnection.AddTrack(t)
 		if err != nil {
 			panic(err)
 		}
--- a/track.go
+++ b/track.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"sync/atomic"
 
 	"github.com/pion/mediadevices/pkg/codec"
 	"github.com/pion/mediadevices/pkg/driver"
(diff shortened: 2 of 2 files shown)


examples/simple/main.go
```diff
@@ -66,6 +66,11 @@ func main() {
 	}
 
 	for _, tracker := range s.GetTracks() {
+		t := tracker.Track()
+		tracker.OnEnded(func(err error) {
+			fmt.Printf("Track (ID: %s, Label: %s) ended with error: %v\n",
+				t.ID(), t.Label(), err)
```
at-wat: It works like:
```
Track (ID: video, Label: 10afd1bd-dee0-452a-85eb-b49cbea60194) ended with error: EOF
```
on camera disconnection.

BTW, ID and Label are swapped?

lherman-cs: Oops, I'll fix that. Here's the issue, https://github.com/pion/mediadevices/issues/32.

examples/simple/main.go
```diff
@@ -66,6 +66,11 @@ func main() {
...
+		})
 		_, err = peerConnection.AddTrack(tracker.Track())
```
lherman-cs: Could you please change this to `_, err = peerConnection.AddTrack(t)`?

at-wat: Fixed!

Issue #31, opened 2020-02-10
Inserting custom image processor


at-wat:
It would be useful if a custom image processor (`func CustomImageProcessor(r video.Reader) video.Reader`) can be inserted between device and codec.

Personally, I would like to use this package as a replacement of gstreamer. Clock overlay to the image is what I want to insert by using it.

lherman-cs:
Definitely agree. In fact, I've designed the driver and codec for this use case. What's missing right now it's just a hook in the middle between device and codec.

PR #34, opened 2020-02-11
Add custom video/audio transforms

lherman-cs:
I had to update the `MediaStreamConstraints` so that the users will be able to use the APIs nicely.

Resolves https://github.com/pion/mediadevices/issues/31.

### DIFF
--- a/examples/transform/main.go
+++ b/examples/transform/main.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"fmt"
+	"image"
+
+	"github.com/pion/mediadevices"
+	"github.com/pion/mediadevices/examples/internal/signal"
+	_ "github.com/pion/mediadevices/pkg/codec/openh264" // This is required to register h264 video encoder
+	_ "github.com/pion/mediadevices/pkg/codec/opus"     // This is required to register opus audio encoder
+	_ "github.com/pion/mediadevices/pkg/codec/vpx"
+	"github.com/pion/mediadevices/pkg/frame"
+	"github.com/pion/mediadevices/pkg/io/video"
+	"github.com/pion/webrtc/v2"
+)
+
+const (
+	videoCodecName = webrtc.VP8
+)
+
+func removeBlue(r video.Reader) video.Reader {
+	return video.ReaderFunc(func() (img image.Image, err error) {
+		img, err = r.Read()
+		if err != nil {
+			return
+		}
+
+		yuvImg, ok := img.(*image.YCbCr)
+		if !ok {
+			return img, nil
+		}
+
+		yuvImg.Cb = make([]uint8, len(yuvImg.Cb))
+		return
+	})
+}
+
+func main() {
+	config := webrtc.Configuration{
+		ICEServers: []webrtc.ICEServer{
+			{
+				URLs: []string{"stun:stun.l.google.com:19302"},
+			},
+		},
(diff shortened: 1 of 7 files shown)


examples/transform/main.go
```diff
@@ -0,0 +1,135 @@
...
+	vb.Enable().WithCodec(prop.Codec{
+		CodecName: videoCodecName,
+		BitRate:   100000, // 100kbps
+	})
+	vb.WithProp(prop.Video{
+		FrameFormat: frame.FormatI420, // most of the encoder accepts I420
+		Width:       640,
+		Height:      480,
+	})
+	vb.WithTransforms(removeBlue)
```
at-wat:
```go
	var vb mediadevices.VideoConstraintsBuilder
	vb.Enable().
		WithCodec(prop.Codec{
			CodecName: videoCodecName,
			BitRate:   100000, // 100kbps
		}).
		WithProp(prop.Video{
			FrameFormat: frame.FormatI420, // most of the encoder accepts I420
			Width:       640,
			Height:      480,
		}).
		WithTransforms(removeBlue)
```
would be bit less redundant?

lherman-cs:
> Personally, I like functional options pattern rather than method chaining.
> But using it looks bit complicated due to the namespace.
> Any thoughts?

Yea.. I definitely agree. I don't honestly like method chaining. But, the problem with functional options though is that the users can specify non-related constraints. For example:

```
Video: func(c *MediaTrackConstraints) {
    c.ChannelCount = 2
}
```

I guess the pros and cons would be:

Functional Options:
Pros:
* Definitely look cleaner on the user side
* Less code on our side

Cons:
* User might be confused with the options that they can set because non-related constraints are also included.
* 

Method Chaining or Builder Pattern:
Pros and Cons are basically just the flip side of Functional Options' Pros and Cons.

So, I guess, we'll keep the `MediaTrackConstraints` to be:

```
type MediaTrackConstraints struct {
	prop.Media
	Enabled        bool
	VideoTransform video.TransformFunc
	AudioTransform audio.TransformFunc
}
```

And, we'll keep functional options. What do you think? Do you like the naming and where `VideoTransform` and `AudioTransform` by the way? I'm curious, because I don't really like it, but I don't have a better idea atm.

examples/transform/main.go
lherman-cs: Ah, yep. That makes sense.

at-wat:
> So, I guess, we'll keep the `MediaTrackConstraints` to be:
> 
> ```
> type MediaTrackConstraints struct {
> 	prop.Media
> 	Enabled        bool
> 	VideoTransform video.TransformFunc
> 	AudioTransform audio.TransformFunc
> }
> ```
> 
> And, we'll keep functional options. What do you think?

Sounds reasonable to me.

> Do you like the naming and where `VideoTransform` and `AudioTransform` by the way?

They are not actually parts of `Constraints`, but I also don't have good alternative.
So, it's fine for now.

lherman-cs:
> > Do you like the naming and where `VideoTransform` and `AudioTransform` by the way?
> 
> They are not actually parts of `Constraints`, but I also don't have good alternative.
> So, it's fine for now.

Yea.. Definitely agree. It looks weird to be a part of `Constraints`

lherman-cs:
Just updated the PR.

Issue #35, opened 2020-02-11
Read from camera times-out on go1.14rc1


at-wat:
It's not a problem for now, but I would like to leave a note.

Following test just checks `OnEnded` callback.
```go
package main

import (
	"testing"
	"time"

	"github.com/pion/mediadevices"
	_ "github.com/pion/mediadevices/pkg/codec/vpx"
	"github.com/pion/mediadevices/pkg/frame"
	"github.com/pion/webrtc/v2"
)

func TestMain(t *testing.T) {
	configs := map[string]webrtc.Configuration{
		"WithSTUN": {
			ICEServers: []webrtc.ICEServer{
				{URLs: []string{"stun:stun.l.google.com:19302"}},
			},
		},
		"WithoutSTUN": {
			ICEServers: []webrtc.ICEServer{},
		},
	}
	for name, config := range configs {
		t.Run(name, func(t *testing.T) {
			peerConnection, err := webrtc.NewPeerConnection(config)
			if err != nil {
				t.Fatal(err)
			}

			md := mediadevices.NewMediaDevices(peerConnection)

			s, err := md.GetUserMedia(mediadevices.MediaStreamConstraints{
				Video: func(c *mediadevices.MediaTrackConstraints) {
					c.CodecName = videoCodecName
					c.FrameFormat = frame.FormatI420
					c.Enabled = true
					c.Width = 640
					c.Height = 480
				},
			})
			if err != nil {
				t.Fatal(err)
			}
			trackers := s.GetTracks()
			if len(trackers) != 1 {
				t.Fatal("wrong number of the tracks")
			}
			peerConnection.AddTrack(trackers[0].Track())
			trackers[0].OnEnded(func(err error) {
				t.Error(err)
			})
			time.Sleep(10 * time.Second)
			trackers[0].OnEnded(func(err error) {})
			peerConnection.Close()
			trackers[0].Stop()
			time.Sleep(time.Second)
		})
	}
}
```
with treating camera read timeout as error:
```diff
diff --git a/pkg/driver/camera/camera_linux.go b/pkg/driver/camera/camera_linux.go
index cee43b2..f7202f8 100644
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -4,6 +4,7 @@ package camera
 import "C"
 
 import (
+       "errors"
        "image"
        "io"
 
[...]

PR #36, opened 2020-02-11
Stop video track safely

at-wat:
Treat camera read timeout as error and return detailed errors.
Wait finishing video read loop to avoid memory violation
as the video frames might be alloced by C or on shared memory.

### Reference issue
https://github.com/pion/mediadevices/issues/35#issuecomment-584535022

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -4,15 +4,26 @@ package camera
 import "C"
 
 import (
+	"errors"
 	"image"
 	"io"
 	"io/ioutil"
+	"sync/atomic"
 
 	"github.com/blackjack/webcam"
 	"github.com/pion/mediadevices/pkg/driver"
 	"github.com/pion/mediadevices/pkg/frame"
 	"github.com/pion/mediadevices/pkg/io/video"
 	"github.com/pion/mediadevices/pkg/prop"
+)
+
+const (
+	maxEmptyFrameCount = 5
+)
+
+var (
+	errReadTimeout = errors.New("read timeout")
+	errEmptyFrame  = errors.New("empty frame")
 )
 
 // Camera implementation using v4l2
@@ -23,6 +34,7 @@ type camera struct {
 	formats         map[webcam.PixelFormat]frame.Format
 	reversedFormats map[frame.Format]webcam.PixelFormat
 	started         bool
+	closed          atomic.Value // bool
 }
 
 func init() {
@@ -50,11 +62,13 @@ func newCamera(path string) *camera {
 		reversedFormats[v] = k
 	}
 
-	return &camera{
+	c := &camera{
 		path:            path,
 		formats:         formats,
 		reversedFormats: reversedFormats,
 	}
+	c.closed.Store(false)
+	return c
 }
 
 func (c *camera) Open() error {
@@ -64,6 +78,7 @@ func (c *camera) Open() error {
 	}
 
 	c.cam = cam
+	c.closed.Store(false)
 	return nil
 }
 
@@ -73,9 +88,13 @@ func (c *camera) Close() error {
 	}
 
 	if c.started {
+		// Note: StopStreaming frees frame buffers even if they are still used in Go code.
+		//       There is currently no convenient way to do this safely.
+		//       So, consumer of this stream must close camera after unusing all images.
 		c.cam.StopStreaming()
 	}
 	c.cam.Close()
+	c.closed.Store(true)
 	return nil
 }
 
@@ -98,30 +117,37 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 
 	r := video.ReaderFunc(func() (img image.Image, err error) {
 		// Wait until a frame is ready
-		for {
-			err := c.cam.WaitForFrame(5)
+		for i := 0; i < maxEmptyFrameCount; i++ {
+			if c.closed.Load().(bool) {
+				// Return EOF if the camera is already closed.
+				return nil, io.EOF
+			}
+
+			err := c.cam.WaitForFrame(5) // 5 seconds
 			switch err.(type) {
 			case nil:
 			case *webcam.Timeout:
-				continue
+				return nil, errReadTimeout
 			default:
 				// Camera has been stopped.
-				return nil, io.EOF
+				return nil, err
 			}
 
 			b, err := c.cam.ReadFrame()
 			if err != nil {
 				// Camera has been stopped.
-				return nil, io.EOF
+				return nil, err
 			}
 
-			// Frame is not ready.
+			// Frame is empty.
+			// Retry reading and return errEmptyFrame if it exceeds maxEmptyFrameCount.
 			if len(b) == 0 {
 				continue
 			}
 
 			return decoder.Decode(b, p.Width, p.Height)
 		}
+		return nil, errEmptyFrame
 	})
 
 	return r, nil
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -35,7 +35,7 @@ type camera struct {
 	formats         map[webcam.PixelFormat]frame.Format
 	reversedFormats map[frame.Format]webcam.PixelFormat
 	started         bool
-	wg              *sync.WaitGroup
+	mutex           sync.Mutex
 	cancel          func()
 }
 
@@ -90,8 +90,9 @@ func (c *camera) Close() error {
 	if c.cancel != nil {
 		// Let the reader knows that the caller has closed the camera
 		c.cancel()
-		// Wait until the reader receives the cancelation
-		c.wg.Wait()
+		// Wait until the reader unref the buffer
+		c.mutex.Lock()
+		defer c.mutex.Unlock()
 
 		// Note: StopStreaming frees frame buffers even if they are still used in Go code.
 		//       There is currently no convenient way to do this safely.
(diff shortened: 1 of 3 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -92,21 +101,21 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 
 	r := video.ReaderFunc(func() (img image.Image, err error) {
 		// Wait until a frame is ready
-		for {
+		for i := 0; i < maxErrorCount; i++ {
```
lherman-cs: I don't think this block will loop over. All the code paths look like they will just return. Am I missing something?

pkg/driver/camera/camera_linux.go
```diff
@@ -92,21 +101,21 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
...
 			err := c.cam.WaitForFrame(5)
 			switch err.(type) {
 			case nil:
 			case *webcam.Timeout:
-				continue
+				return nil, errReadTimeout
 			default:
 				// Camera has been stopped.
-				return nil, io.EOF
+				return nil, err
```
lherman-cs: Sadly, we somehow need to return `io.EOF` because this is a way for the reader consumer to know when a reader is done emitting values. Essentially, I'm trying to follow `io.Reader` rules.

pkg/driver/camera/camera_linux.go
```diff
@@ -92,21 +101,21 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
...
 	r := video.ReaderFunc(func() (img image.Image, err error) {
 		// Wait until a frame is ready
-		for {
+		for i := 0; i < maxErrorCount; i++ {
 			err := c.cam.WaitForFrame(5)
 			switch err.(type) {
 			case nil:
 			case *webcam.Timeout:
-				continue
+				return nil, errReadTimeout
```
lherman-cs: Is this supposed to be `continue` maybe?

pkg/driver/camera/camera_linux.go
lherman-cs: Btw, how limiting the loop helps with the segfault? Because it looks to me that the synchronization that you did in video track already prevented it. I feel like when there's a timeout, it's better for the reader to keep looping until it gets a frame otherwise the reader consumer will think that there's an actual error and simply stop consuming. What do you think?

at-wat: This is intended to count continue at:
```go
			if len(b) == 0 {
				continue
			}
```
which must be a kind of fatal error. Naming of the const might be confusing.

at-wat: > I feel like when there's a timeout, it's better for the reader to keep looping until it gets a frame otherwise the reader consumer will think that there's an actual error and simply stop consuming.

The frame buffer returned from webcam package is on Go heap but managed by mmap, and unref-ed on `stopStreaming`. 
(Badly, webcam package stops reading and immediately unref the buffer. GC may immediately release the buffer even if it is used by the consumer.)
If the consumer still accesses this `img image.Image` after calling `stopStreaming`, it causes access violation.
After returning error, it can be assumed that the buffer returned previously is no longer accessed.
So, if this loop can live long, this driver can't be safely stopped.

pkg/driver/camera/camera_linux.go
at-wat: Umm, in my feeling, 5 seconds of the timeout is long enough to treat as an error.

pkg/driver/camera/camera_linux.go
at-wat: In my understanding, `io.Reader` returns `io.EOF` only if reaches the end of the data and has no error, and consumer can't assume that the reader resumes after any error.
Please let me know if my understanding is wrong.

pkg/driver/camera/camera_linux.go
lherman-cs: > This is intended to count continue at:
> 
> ```go
> 			if len(b) == 0 {
> 				continue
> 			}
> ```
> 
> which must be a kind of fatal error. Naming of the const might be confusing.

Lol, apparently, Github cut off the snippet right when there's a statement to continue...

This is what I'm seeing
Screenshot from 2020-02-11 21-25-16

lherman-cs: Right. But, your synchronization in video track already makes sure that there will be no call to this Reader anymore and this reader will only be closed until the loop in the video track stops. So, technically, nobody will read the image which can segfault since this image never gets emitted in the first place.

Btw, I feel like this problem should be solved in `drive` package itself. I feel weird to let video track to handle this kind of implicit scenario. I wonder if we can apply the synchronization that you did in video track in `wrapper.go` instead.

pkg/driver/camera/camera_linux.go
lherman-cs: Oops. My bad. That makes sense. Yes, 5 seconds is way too long... It should be an error.

pkg/driver/camera/camera_linux.go
lherman-cs: Yep, that's definitely true. I think my wording was just really bad. `io.EOF` is definitely like any error, except it adds value to the caller that this is one of the error values which they can ignore (since it's not really an error but rather just a way to tell the caller that it's the end of file).

I guess the difference is that if there's another function that takes this reader, it can ignore `io.EOF` and still return `nil` for its own error.

A good example is `io.Copy` which under the hood calls `copyBuffer`, https://golang.org/src/io/io.go?s=13434:13516#L380. In there, you can see that it treats `io.EOF` as non-error.

pkg/driver/camera/camera_linux.go
at-wat: > Btw, I feel like this problem should be solved in drive package itself

I agree, and I tried to do so at first.
But, due to the implementation of `webcam.StopStreaming`, `StopStreaming` must be called after the consumer unref the buffer, which could not be known by the driver.
Only case, that the driver can assume that the previously returned buffer is unreffed, is returning an error from Read. (Sorry, https://github.com/pion/mediadevices/pull/36#discussion_r378042023 mixed this thought.)

pkg/driver/camera/camera_linux.go
at-wat: Yes, but I think it's better to notify device disconnection or IO failure kind of errors as error.
I digged into webcam package now, and found that `WaitForFrame` and `ReadFrame` only returns error fom `unix.Select` and `unix.Syscall`, they basically a kind of fatal.

I think it would be better to return `io.EOF` if `Read` is called after `Close`.

at-wat: > I think it would be better to return `io.EOF` if `Read` is called after `Close`.

I'll add this.

at-wat:
Added some more comments and fix const name.
`Read` now returns `io.EOF` if called after `Close`.

pkg/driver/camera/camera_linux.go
lherman-cs: How about if we create a buffer that contains the frame before we decode it, essentially moving the memory from mmap to Go heap. It definitely requires a copy every time we get a frame. But, this will guarantee that all the data that comes from this Reader will be Go safe since the memory now is managed by Go. What do you think?

lherman-cs: I've created a POC for the idea, https://github.com/pion/mediadevices/blob/stop-video-track-safely-exp/pkg/driver/camera/camera_linux.go. I borrowed most of your synchronization that you did in video track (basically moved it to `camera_linux.go`), and added the intermediate buffer that will be used to the caller. Please let me know what you think about it.

at-wat: stop-video-track-safely-exp branch seems not been pushed?
I'm curious about difference of the performance. If there is no significant change, I definitely agree to add a buffer.

lherman-cs: Lol, sorry. I forgot to commit... And just pushed it. I've updated it. It doesn't seem that bad, honestly.

at-wat: I think if we copy the buffer, WaitGroup is no longer required as the buffer is managed under Go GC.

at-wat: Added CPU usage bench to stop-video-track-safely branch.
The difference of CPU usage is very small and difficult to be observed on my environment.
I absolutely agree to copy the buffer to Go heap.

Could you force push it to this branch?

lherman-cs: > I think if we copy the buffer, WaitGroup is no longer required as the buffer is managed under Go GC.

It's probably not required. But, in my head, there's a race condition where `StopStreaming` is being called (from a different goroutine) right before we copy. That means that it can cause a segfault during the copy since we're still accessing a freed memory. Honestly, it should be super rare, definitely an edge case. What do you think? I might be wrong...

at-wat: Oh, you are right.
How about adding a lock at the beginning of ReaderFunc
https://github.com/pion/mediadevices/blob/2ec158e581f5032d59b089703ffdfc85cc671495/pkg/driver/camera/camera_linux.go#L121-L123
and `StopStreaming()`
https://github.com/pion/mediadevices/blob/2ec158e581f5032d59b089703ffdfc85cc671495/pkg/driver/camera/camera_linux.go#L93
As the context is already canceled when passing the lock on ReaderFunc, it will just return `io.EOF`.

lherman-cs: Hmm, that sounds reasonable to me. That way, we can avoid the deadlock problem. The only concern I have is that every time we read, we need to acquire the lock. But, hopefully, the performance penalty is minuscule.

at-wat: Added a commit for this!

PR #40, opened 2020-02-12
Implement EnumerateDevices and DeviceID property

### LINKED ISSUE #38: Filter device by deviceId
It is `MediaTrackConstraints.deviceId` in Web API.

at-wat:
Implement EnumerateDevices and register all devices
Select media with DeviceID property

### Reference issue
Closes #38

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/blackjack/webcam v0.0.0-20191123110216-08fa32efcb67
 	github.com/faiface/beep v1.0.2
-	github.com/jfreymuth/pulse v0.0.0-20200107133239-fe42f62ea140
+	github.com/jfreymuth/pulse v0.0.0-20200118113426-7cf5f487291e
 	github.com/pion/webrtc/v2 v2.1.19-0.20200106051345-726a16faa60d
 	github.com/satori/go.uuid v1.2.0
 	gopkg.in/hraban/opus.v2 v2.0.0-20191117073431-57179dff69a6
--- a/mediadevices.go
+++ b/mediadevices.go
@@ -9,6 +9,8 @@ import (
 	"github.com/pion/webrtc/v2"
 )
 
+var errNotFound = fmt.Errorf("failed to find the best driver that fits the constraints")
+
 // MediaDevices is an interface that's defined on https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices
 type MediaDevices interface {
 	GetUserMedia(constraints MediaStreamConstraints) (MediaStream, error)
@@ -130,7 +132,7 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
 	}
 
 	if bestDriver == nil {
-		return nil, MediaTrackConstraints{}, fmt.Errorf("failed to find the best driver that fits the constraints")
+		return nil, MediaTrackConstraints{}, errNotFound
 	}
 
 	// Reset Codec because bestProp only contains either audio.Prop or video.Prop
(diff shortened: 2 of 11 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -25,11 +26,16 @@ type camera struct {
...
-	driver.GetManager().Register(defaultCam)
+	searchPath := "/dev/v4l/by-path/"
+	devices, err := ioutil.ReadDir(searchPath)
+	if err != nil {
+		// No v4l device.
+		return
+	}
+	for _, device := range devices {
+		cam := newCamera(searchPath + device.Name())
+		driver.GetManager().Register(cam, device.Name())
```
at-wat: It's impossible to determine the camera label like `front camera` on Linux.
Number in `/dev/video*` and `/dev/v4l/by-id/` depends on the timing of device plug-in.
For these reasons, I put device path (for example `pci-0000\:00\:14.0-usb-0\:9.3\:1.0-video-index0`), which is unique according to the USB port, as a label.

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.13
 require (
 	github.com/blackjack/webcam v0.0.0-20191123110216-08fa32efcb67
 	github.com/faiface/beep v1.0.2
-	github.com/jfreymuth/pulse v0.0.0-20200107133239-fe42f62ea140
```
at-wat: Old version had a bug on device selection by ID.

mediadevices.go
```diff
@@ -142,18 +145,55 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
 }
 
 func (m *mediaDevices) selectAudio(constraints MediaTrackConstraints) (Tracker, error) {
-	d, c, err := selectBestDriver(driver.FilterAudioRecorder(), constraints)
+	filter := driver.FilterAudioRecorder()
+	if constraints.DeviceID != "" {
+		filter = func(d driver.Driver) bool {
+			return driver.FilterAudioRecorder()(d) && driver.FilterID(constraints.DeviceID)(d)
```
lherman-cs: I think it's more readable if it's written like the following,

```
audioFilter := driver.FilterAudioRecorder() // or even just filter
idFilter := driver.FilterID(constraints.DeviceID)
filter = func(d driver.Driver) bool {
	return audioFilter(d) && idFilter(d)
```

But, honestly, I don't have a hard feeling on this, just giving an opinion ^^. Feel free to ignore it.

mediadevices.go
```diff
@@ -142,18 +145,55 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
...
 		return nil, err
 	}
 
 	return newAudioTrack(m.codecs[webrtc.RTPCodecTypeAudio], d, c)
 }
 func (m *mediaDevices) selectVideo(constraints MediaTrackConstraints) (Tracker, error) {
-	d, c, err := selectBestDriver(driver.FilterVideoRecorder(), constraints)
+	filter := driver.FilterVideoRecorder()
+	if constraints.DeviceID != "" {
+		filter = func(d driver.Driver) bool {
```
lherman-cs: Same as above.

mediadevices.go
```diff
@@ -142,18 +145,55 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
...
 	return newVideoTrack(m.codecs[webrtc.RTPCodecTypeVideo], d, c)
 }
+
+func (m *mediaDevices) EnumerateDevices() []MediaDeviceInfo {
+	drivers := driver.GetManager().Query(
+		driver.FilterFn(func(driver.Driver) bool { return true }))
+	info := make([]MediaDeviceInfo, 0, len(drivers))
+	for _, d := range drivers {
+		var kind MediaDeviceType
+		switch {
```
lherman-cs: I like this. It makes sense. Instead of using type inference, we can use these functions to hide that. Pretty cool!

pkg/driver/camera/camera_linux.go
lherman-cs: I'm about this. On my laptop, I have 2 registered cameras, `/dev/video0` and `/dev/video1`. But, `/dev/video1` is basically not usable, I couldn't even get its information using v4l cli. Do you know why?

pkg/driver/manager.go
```diff
@@ -20,6 +20,13 @@ func FilterAudioRecorder() FilterFn {
 	}
 }
 
+// FilterID return a filter function to get registered AudioRecorder which have given ID
```
lherman-cs: "*get registered drivers" since it's not only `AudioRecorder`

pkg/driver/camera/camera_linux.go
at-wat: It is reported on Kernel.org Bugzilla.
https://bugzilla.kernel.org/show_bug.cgi?id=199575
Second one seems for taking metadata of the camera.

at-wat:
Comment and filtering part are fixed. PTAL

pkg/driver/camera/camera_linux.go
lherman-cs: Ah, I see. Thanks ^^

PR #42, opened 2020-02-13
Add screen capture device for Linux

### LINKED ISSUE #20: Add screen capture adapter for Linux
For linux, libx11 `XOpenDisplay, XDefaultRootWindow, XShmGetImage, ShmPutImage` could be used.

at-wat:
It works like:
Screenshot_2020-02-13 Edit fiddle - JSFiddle

### Reference issue
Fixes #20

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -8,6 +8,6 @@ import (
 )
 
 // RegisterDriverAdapter allows user space level of driver registration
-func RegisterDriverAdapter(a driver.Adapter, label string) error {
-	return driver.GetManager().Register(a, label)
+func RegisterDriverAdapter(a driver.Adapter, info driver.Info) error {
+	return driver.GetManager().Register(a, info)
 }
--- a/driver.go
+++ b/driver.go
@@ -4,6 +4,7 @@ import (
 	"github.com/pion/mediadevices/pkg/driver"
 	_ "github.com/pion/mediadevices/pkg/driver/camera"
 	_ "github.com/pion/mediadevices/pkg/driver/microphone"
+	_ "github.com/pion/mediadevices/pkg/driver/screen"
 )
 
 // RegisterDriverAdapter allows user space level of driver registration
--- a/mediadevices.go
+++ b/mediadevices.go
@@ -13,6 +13,7 @@ var errNotFound = fmt.Errorf("failed to find the best driver that fits the const
 
 // MediaDevices is an interface that's defined on https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices
 type MediaDevices interface {
+	GetDisplayMedia(constraints MediaStreamConstraints) (MediaStream, error)
 	GetUserMedia(constraints MediaStreamConstraints) (MediaStream, error)
 	EnumerateDevices() []MediaDeviceInfo
 }
@@ -73,6 +74,34 @@ func WithTrackGenerator(gen TrackGenerator) MediaDevicesOption {
 	return func(o *MediaDevicesOptions) {
 		o.trackGenerator = gen
 	}
+}
+
+// GetDisplayMedia prompts the user to select and grant permission to capture the contents
+// of a display or portion thereof (such as a window) as a MediaStream.
+// Reference: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia
+func (m *mediaDevices) GetDisplayMedia(constraints MediaStreamConstraints) (MediaStream, error) {
+	trackers := make([]Tracker, 0)
+
+	var videoConstraints MediaTrackConstraints
+	if constraints.Video != nil {
+		constraints.Video(&videoConstraints)
+	}
+
+	if videoConstraints.Enabled {
+		tracker, err := m.selectScreen(videoConstraints)
+		if err != nil {
+			return nil, err
+		}
+
+		trackers = append(trackers, tracker)
+	}
+
+	s, err := NewMediaStream(trackers...)
+	if err != nil {
+		return nil, err
+	}
+
+	return s, nil
 }
 
 // GetUserMedia prompts the user for permission to use a media input which produces a MediaStream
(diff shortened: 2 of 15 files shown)


at-wat:
It's quite heavy on full HD screen. A transform for downsampling will be needed.

driver.go
```diff
@@ -4,6 +4,7 @@ import (
 	"github.com/pion/mediadevices/pkg/driver"
 	_ "github.com/pion/mediadevices/pkg/driver/camera"
 	_ "github.com/pion/mediadevices/pkg/driver/microphone"
+	_ "github.com/pion/mediadevices/pkg/driver/screen"
```
at-wat: BTW, it maybe better to load drivers by user code as same as codecs to minimize the binary size and number of dependencies according to the specific usage.

lherman-cs: I actually thought about that. But, I wasn't sure if it was worth it. Because if we do it like this, we'll add some complexities to the users, they need to know what to import for the getUserMedia to work properly which can be somewhat unintuitive especially for Go beginners.

lherman-cs: You actually brought some good points here, binary size and number of dependencies. Love it! I'll create a PR for this.

pkg/driver/screen/x11_linux.go
```diff
@@ -0,0 +1,107 @@
...
+		return imgI444, nil
+	})
+	return r, nil
+}
+
+func (s *screen) Properties() []prop.Media {
+	rect := s.reader.img.Bounds()
+	w := rect.Max.X - rect.Min.X
+	h := rect.Max.Y - rect.Min.Y
+	return []prop.Media{
```
lherman-cs: It looks like we need extra information to let us know that an adapter can be used for screen sharing. I saw that there's a web API called, https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia. It's very similar to `getMediaDevices` but for screen sharing devices.

I've created a POC branch that implements `getDisplayMedia`, https://github.com/pion/mediadevices/tree/GetDisplayMedia. Please let me know what you think about it.

POC Overview:

    Change driver registration requirement to have an Info
    
    * Move label to Info
    * Replace label requirement to be an info
    * Replace Label driver method to be an Info method

    Add DeviceType to Info and a filter function for it
    
    * Add DeviceType to Info
    * Update adapters to also register its device type
    * Add FilterDeviceType

    Add GetDisplayMedia
    
    * Add selectScreen to select all registered drivers that can screenshare
    * Add GetDisplayMedia
    * Add screenshare example

at-wat: Oh, I didn't know `getDisplayMedia`.

On browser, `enumerateDevice` seems not listing screen capture devices, and I guess `getUserMedia` never select screen capture device.
Maybe it's better to omit screen capture from our `GetDisplayMedia`?
(But, enumerating screens is helpful.)

Please feel free to push to this branch.

lherman-cs: Hmm, that's interesting. I agree I think we should omit screen capture devices from `getUserMedia`. 

btw, are you suggesting that we should still keep the screen capture devices for `EnumerateDevices`?

at-wat: > btw, are you suggesting that we should still keep the screen capture devices for EnumerateDevices?

I'm not very sure it's good, but there may be multiple screens the user can select.
At least, there should be a way to get the number of the screens.

lherman-cs: Sadly, the spec for `enumerateDevices` doesn't say anything about screen capture devices, it only talks about camera and microphone. https://w3c.github.io/mediacapture-main/#dom-mediadevices-enumeratedevices.

lherman-cs: > > btw, are you suggesting that we should still keep the screen capture devices for EnumerateDevices?
> 
> I'm not very sure it's good, but there may be multiple screens the user can select.
> At least, there should be a way to get the number of the screens.

That's definitely a good point. Should we just expose `DeviceType` in `EnumerateDevices`. it's kinda weird that it doesn't follow the spec. But, I think it's pretty clean. What do you think?

at-wat: > Should we just expose DeviceType in EnumerateDevices. 

Totally agree with this!

lherman-cs: Sweet. I'll add this then.

pkg/driver/manager.go
```diff
@@ -27,6 +27,13 @@ func FilterID(id string) FilterFn {
 	}
 }
 
+// FilterDeviceType returns a filter function to get registered drivers which matches t type
+func FilterDeviceType(t DeviceType) FilterFn {
+	return func(d Driver) bool {
+		return d.Info().DeviceType == t
+	}
+}
+
```
at-wat: It maybe nice to have
```go
// FilterAnd returns a filter function to take logical conjunction of given filters.
func FilterAnd(filters... FilterFn) FilterFn {
	return func(d Driver) bool {
		for _, f := range filters {
			if !f(d) {
				return false
			}
		}
		return true
	}
}
```

pkg/driver/screen/x11_linux.go
lherman-cs: added

pkg/driver/manager.go
lherman-cs: Oh yes, this looks really good! Feel free to add it. This will make the code cleaner for sure.

lherman-cs:
> It's quite heavy on full HD screen. A transform for downsampling will be needed.

I'm thinking to add a bunch of helper transformers under `io/video`. Downsampling can definitely be one of them!

pkg/driver/manager.go
at-wat: I'll add it!

at-wat: Added

mediadevices.go
```diff
@@ -195,13 +222,29 @@ func (m *mediaDevices) selectAudio(constraints MediaTrackConstraints) (Tracker,
 	return newAudioTrack(&m.MediaDevicesOptions, d, c)
 }
 func (m *mediaDevices) selectVideo(constraints MediaTrackConstraints) (Tracker, error) {
-	filter := driver.FilterVideoRecorder()
+	typeFilter := driver.FilterVideoRecorder()
+	screenFilter := driver.FilterDeviceType(driver.Screen)
+	filter := driver.FilterAnd(typeFilter, screenFilter)
```
at-wat: not(screen)

PR #48, opened 2020-02-14
Add audiotest and videotest driver

at-wat:
Which work similar to GStreamer audiotestsrc, videotestsrc.

image

### DIFF
--- a/pkg/driver/audiotest/dummy.go
+++ b/pkg/driver/audiotest/dummy.go
@@ -1,0 +1,90 @@
+// Package audiotest provides dummy audio driver for testing.
+package audiotest
+
+import (
+	"context"
+	"io"
+	"math"
+	"time"
+
+	"github.com/pion/mediadevices/pkg/driver"
+	"github.com/pion/mediadevices/pkg/io/audio"
+	"github.com/pion/mediadevices/pkg/prop"
+)
+
+func init() {
+	driver.GetManager().Register(
+		&dummy{}, driver.Info{Label: "AudioTest", DeviceType: driver.Microphone},
+	)
+}
+
+type dummy struct {
+	closed <-chan struct{}
+	cancel func()
+}
+
+func (d *dummy) Open() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	d.closed = ctx.Done()
+	d.cancel = cancel
+	return nil
+}
+
+func (d *dummy) Close() error {
+	d.cancel()
+	return nil
+}
+
+func (d *dummy) AudioRecord(p prop.Media) (audio.Reader, error) {
+	var sin [100]float32
+	for i := range sin {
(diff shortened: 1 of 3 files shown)


pkg/driver/videotest/dummy.go
```diff
@@ -0,0 +1,115 @@
...
+		{107, 202, 222},
+		{82, 90, 240},
+		{41, 240, 110},
+	}
+
+	r := video.ReaderFunc(func() (image.Image, error) {
+		select {
+		case <-d.closed:
+			return nil, io.EOF
+		default:
```
at-wat: need to add ticker

at-wat:
It may be nice to add an instruction to use this in the examples, for users who don't have camera right now.

at-wat:
rebased

pkg/driver/videotest/dummy.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/pion/mediadevices/pkg/driver"
+	"github.com/pion/mediadevices/pkg/frame"
+	"github.com/pion/mediadevices/pkg/io/video"
+	"github.com/pion/mediadevices/pkg/prop"
+)
+
+func init() {
+	driver.GetManager().Register(
+		newVideoTest(),
+		driver.Info{Label: "VideoTest", DeviceType: driver.Microphone},
```
lherman-cs: Oops, it should be `driver.Camera` there

pkg/driver/audiotest/dummy.go
```diff
@@ -0,0 +1,90 @@
...
+func init() {
+	driver.GetManager().Register(
+		&dummy{}, driver.Info{Label: "AudioTest", DeviceType: driver.Microphone},
+	)
+}
+
+type dummy struct {
+	closed <-chan struct{}
+	cancel func()
+	phase  int
```
lherman-cs: phase can be just inside AudioRecord, less scope.

pkg/driver/videotest/dummy.go
at-wat: oops

lherman-cs:
> It may be nice to add an instruction to use this in the examples, for users who don't have camera right now.

That's a good point. How about if we just add a comment in the simple example, above the camera and microphone imports?

Something like:
```
	// Note: If you don't have a camera or microphone or your adapters are not supported,
	//_ "github.com/pion/mediadevices/pkg/driver/audiotest"
	_ "github.com/pion/mediadevices/pkg/driver/camera"     // This is required to register camera adapter
	_ "github.com/pion/mediadevices/pkg/driver/microphone" // This is required to register microphone adapte
```

at-wat:
https://github.com/pion/mediadevices/pull/48#issuecomment-586555954 sounds very nice:+1:

at-wat:
Addressed you comments! I'll rebase and add the comment to the simple example.

PR #54, opened 2020-02-15
Add support for non-YCbCr input to ToI420

lherman-cs:
Resolves https://github.com/pion/mediadevices/issues/46

PR #55, opened 2020-02-15
Add video scaler

### LINKED ISSUE #53: Add video scaling transform
Some scaling kernels are available in `x/image/draw` package.
https://pkg.go.dev/golang.org/x/image/draw?tab=doc#pkg-variables

at-wat:
- Add video scaler transform
- Add scaler to screencast example
- Support size variable encoding of vpx

### Reference issue
Fixes #53


pkg/io/video/scale.go
```diff
@@ -0,0 +1,151 @@
...
+					planes[2].src.Rect.Max.X /= 2
+					planes[2].src.Rect.Max.Y /= 2
+				}
+				// TODO: add option to enable multi threading
+				var wg sync.WaitGroup
+				wg.Add(3)
+				for i, p := range planes {
+					i, p := i, p
+					go func() {
+						scalers[i].Scale(p.dst, p.dst.Rect, p.src, p.src.Rect, draw.Over, nil)
```
at-wat: Scaling Gray image using `x/image/draw` is 10 times slower than RGBA:thinking:

at-wat: Maybe due to Gray->RGBA->Gray color transform?

at-wat: Scaling 1080p to 360p takes:
- YCbCr as 3 `image.Gray`: 60 ms
- YCbCr wrapped by RGBA like image interface: 20ms
- RGBA: 2ms

(`x/image/draw` has optimized implementation only for RGBA to RGBA)

at-wat: 1080p RGBA to I444 is now taking 150ms/frame.
After merging #54, screencast example will scale RGBA image, and transform color of downsampled image. So, the performance will be significantly improved.

examples/screenshare/main.go
```diff
@@ -52,11 +52,10 @@ func main() {
 	s, err := md.GetDisplayMedia(mediadevices.MediaStreamConstraints{
 		Video: func(c *mediadevices.MediaTrackConstraints) {
 			c.CodecName = videoCodecName
-			c.FrameFormat = frame.FormatI444
 			c.Enabled = true
-			c.Width = 640
```
at-wat: Screen adapter is totally passive to the screen size and frame format.

lherman-cs:
This is probably a little bit late. Have you heard about caire project, https://github.com/esimov/caire? It's the same guy who created pigo. The scaling is content-aware.

at-wat:
I just search around the source code of caire, but it looks supporting YCbCr only in `imgToNRGBA` conversion. Any other I/O are NRGBA.

pkg/io/video/scaler_cgo.go
```diff
@@ -0,0 +1,71 @@
...
+
+// List of scaling algorithms
+var (
+	ScalerFastNearestNeighbor = Scaler(&FastNearestNeighbor{})
+)
+
+type FastNearestNeighbor struct {
+}
+
+func (f *FastNearestNeighbor) Scale(dst draw.Image, dr image.Rectangle, src image.Image, sr image.Rectangle, op draw.Op, opts *draw.Options) {
```
at-wat:
```
BenchmarkScale/ApproxBiLinear/1080p/I444-4             	      57	  58631034 ns/op	 3725235 B/op	  921601 allocs/op
BenchmarkScale/ApproxBiLinear/1080p/RGBA-4             	     296	  12010046 ns/op	    6258 B/op	       0 allocs/op
BenchmarkScale/ApproxBiLinear/480p/I444-4              	      57	  58357296 ns/op	 3725244 B/op	  921601 allocs/op
BenchmarkScale/ApproxBiLinear/480p/RGBA-4              	     300	  11671893 ns/op	    6174 B/op	       0 allocs/op
BenchmarkScale/BiLinear/1080p/I444-4                   	      13	 246367621 ns/op	15697491 B/op	 3453844 allocs/op
BenchmarkScale/BiLinear/1080p/RGBA-4                   	      45	  78793023 ns/op	  535641 B/op	       0 allocs/op
BenchmarkScale/BiLinear/480p/I444-4                    	      49	  63338238 ns/op	 3008694 B/op	  690242 allocs/op
BenchmarkScale/BiLinear/480p/RGBA-4                    	     157	  23301369 ns/op	   74981 B/op	       0 allocs/op
BenchmarkScale/FastNearestNeighbor/1080p/I444-4        	     645	   5426510 ns/op	    3430 B/op	       0 allocs/op
BenchmarkScale/FastNearestNeighbor/1080p/RGBA-4        	    2112	   1672352 ns/op	     877 B/op	       0 allocs/op
BenchmarkScale/FastNearestNeighbor/480p/I444-4         	     673	   5189816 ns/op	    3288 B/op	       0 allocs/op
BenchmarkScale/FastNearestNeighbor/480p/RGBA-4         	    2678	   1503768 ns/op	     691 B/op	       0 allocs/op
BenchmarkScale/NearestNeighbor/1080p/I444-4            	     180	  19770705 ns/op	  933903 B/op	  230401 allocs/op
BenchmarkScale/NearestNeighbor/1080p/RGBA-4            	     644	   5476329 ns/op	    2876 B/op	       0 allocs/op
BenchmarkScale/NearestNeighbor/480p/I444-4             	     181	  19865321 ns/op	  933848 B/op	  230401 allocs/op
BenchmarkScale/NearestNeighbor/480p/RGBA-4             	     636	   5347700 ns/op	    2912 B/op	       0 allocs/op
```

lherman-cs: Wow, I love this benchmark

pkg/io/video/scaleycrcb.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (p *rgbLikeYCbCr) Bounds() image.Rectangle {
+	return p.y.Rect
+}
+
+func (p *rgbLikeYCbCr) At(x, y int) color.Color {
+	var yy, cb, cr uint8
+	yy = p.y.GrayAt(x, y).Y
+	if (image.Point{x, y}.In(p.cb.Rect)) {
```
lherman-cs: I think we'll lose some important pixels. It's better to scale down the x and y accordingly for chroma planes. 

Something like,

```
// probably store these in struct instead
x = x * p.y.Dx() / p.cb.Dx()
y = y * p.y.Dy() / p.cb.Dy() 

cb = p.cb.GrayAt(x, y).Y
cr = p.cr.GrayAt(x, y).Y
```
What do you think?

pkg/io/video/scale.go
```diff
@@ -0,0 +1,175 @@
...
+					w := img.Bounds().Dx() * height / img.Bounds().Dy()
+					rect = image.Rect(0, 0, w, height)
+				}
+				cacheScaler(rect, img.Bounds())
+			}
+
+			switch v := img.(type) {
+			case *image.RGBA:
+				rgbaRealloc(v)
+				dst := imgScaled.(*image.RGBA)
```
lherman-cs: Not sure if this is going to happen. But, technically, this can cause a panic if suddenly the stream emits RGBA then later emits YCbCr. What do you think?

pkg/io/video/scale.go
```diff
@@ -0,0 +1,175 @@
...
+	if scaler != nil {
+		scalerCached = scaler
+	}
+	cacheScaler := func(dRect, sRect image.Rectangle) {
+		if kernel, ok := scaler.(interface {
+			NewScaler(int, int, int, int) draw.Scaler
+		}); ok {
+			scalerCached = kernel.NewScaler(dRect.Dx(), dRect.Dy(), sRect.Dx(), sRect.Dy())
+		}
+	}
```
lherman-cs: Rather than checking scaler interface inside of the func, it can be moved outside, and inside the function, it can just be

```
scalerCached = kernel.NewScaler(dRect.Dx(), dRect.Dy(), sRect.Dx(), sRect.Dy())
```

pkg/io/video/scale.go
```diff
@@ -0,0 +1,175 @@
...
+			if imgScaled == nil {
+				if height <= 0 {
+					h := img.Bounds().Dy() * width / img.Bounds().Dx()
+					rect = image.Rect(0, 0, width, h)
+				} else if width <= 0 {
+					w := img.Bounds().Dx() * height / img.Bounds().Dy()
+					rect = image.Rect(0, 0, w, height)
+				}
+				cacheScaler(rect, img.Bounds())
+			}
```
lherman-cs: This might be rare. But, if the image from the reader got resized after the first read, `scalerCached` is no longer valid.

pkg/io/video/scale.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+				scalerCached.Scale(dst, dst.Bounds(), src, src.Bounds(), draw.Src, nil)
+
+				*(imgScaledCopy.(*image.YCbCr)) = *(imgScaled.(*image.YCbCr)) // Clone metadata
+
+			default:
+				return nil, errUnsupportedImageType
+			}
+
+			return imgScaledCopy, nil
```
lherman-cs: I'm curious. Why didn't you return `imgScaled` instead?

at-wat: It's necessary to set metadata each time as the next reader like ToI420 might overwrite it, but it is bit complexed since there are variety of pixel format and sub-sampling. So, I just copy the container (buffer of the slices are shared).

pkg/io/video/scaleycrcb.go
at-wat: It is used to process three gray images (which are the planes of a YCbCr image) at once by one scaling factor, without having overhead of `Gray->RGBA->Gray` color transform in `image/draw`'s Scaler.
(`image.YCbCr` don't have `Set()`, so it can't be directly passed to `Scaler.Scale()`.)
Pixel buffer of `p.cb` and `p.cr` are backended directly by original `YCbCr`'s `Cb`, `Cr` buffer (at [here](https://github.com/pion/mediadevices/pull/55/files#diff-004d7e97caf7d4018cb1de5c46028d98R157-R163)) which actually have smaller size, so all raw pixels in Cb/Cr layers are used.

pkg/io/video/scale.go
at-wat: Right, it and also https://github.com/pion/mediadevices/pull/55/files#r379957449 can happen. 
I'll put format change handling to `rgbaRealloc/ycbcrRealloc`.

at-wat:
Now it handles input image size/format change.

pkg/io/video/scaleycrcb.go
lherman-cs: Right, since Cb/Cr layers are smaller, the if statement will be false if we try to go beyond Cb/Cr sizes but still lower than Y size. 

Let say we have a 16x16 image and a 4:2:2 subsampling, we should be able to say `At(15, 15)` and get the Cb/Cr values at 15,15. But, since the Cb/Cr rectangles are just 16x8, the if statement will be false.

pkg/io/video/scale.go
lherman-cs: Ah I see. If that's the case, then I would not declare `imgScaledCopy`, but just return in the switch case.

So, instead of,

```
*(imgScaledCopy.(*image.RGBA)) = *dst // Clone metadata
```

```
cloned := *dst // Clone metadata
return cloned, nil
```

I think it'll be as performant still, but it's more readable and simpler. What do you think?

pkg/io/video/scaleycrcb.go
at-wat: This image is not handled as YCbCr, but independent 3 Gray images.
For example of 4x4-pixels 4:2:2 image scaled by 1/2 by nearest neighbor,
input image:
```
Y 4x4
1 1 0 0
1 1 0 0
0 0 2 2
0 0 2 2
Cb 4x2
4 4 2 2
4 4 2 2
Cb 4x2
6 6 3 3
6 6 3 3
```
intermediate rgbLikeYCbCr:
```
R 4x4
1 1 0 0
1 1 0 0
0 0 2 2
0 0 2 2
G 4x4
4 4 2 2
4 4 2 2
P P P P // virtually padded, actually out of the capacity
P P P P
B 4x4
6 6 3 3
6 6 3 3
P P P P
P P P P
```
Scaled rgbLikeYCbCr:
```
R 2x2
1 0
0 2
G 2x2
4 2
P P
B 2x2
6 3
P P
```
Reconstructed YCbCr:
```
Y 2x2
1 0
0 2
Cb 2x1
4 2
Cr 2x1
6 3
```

lherman-cs: Ah, I got you. The paddings are not actually there. But, we virtually created them by checking if they're in the boundaries. If not, then we'll just return 0 (padding value).

pkg/io/video/scale.go
at-wat: I think allocation degrades performance a little bit, but looks still acceptable amount.

PR #62, opened 2020-02-16
Optimize image processing

at-wat:
branched from #54 non-ycbcr-input

### DIFF
--- a/pkg/driver/screen/x11capture_linux.go
+++ b/pkg/driver/screen/x11capture_linux.go
@@ -1,11 +1,36 @@
 package screen
 
 // #cgo pkg-config: x11 xext
+// #include <stdint.h>
 // #include <sys/shm.h>
 // #include <X11/Xlib.h>
 // #define XUTIL_DEFINE_FUNCTIONS
 // #include <X11/Xutil.h>
 // #include <X11/extensions/XShm.h>
+// void copyRGBA(void *dst, char *src, size_t l) { // 64bit aligned copy
+//   uint64_t *d = (uint64_t*)dst;
+//   uint64_t *s = (uint64_t*)src;
+//   l /= 8;
+//   for (size_t i = 0; i < l; i ++) {
+//     uint64_t v = *s;
+//     // Reorder BGR to RGB
+//     *d = 0xFF000000FF000000 |
+//          ((v >> 16) & 0xFF00000000) | (v & 0xFF0000000000) | ((v & 0xFF00000000) << 16) |
+//          ((v >> 16) & 0xFF) | (v & 0xFF00) | ((v & 0xFF) << 16);
+//     d++;
+//     s++;
+//   }
+// }
+// char *align64(char *ptr) { // return 64bit aligned pointer
+//   if (((size_t)ptr & 0x07) == 0) {
+//     return ptr;
+//   }
+//   // Clear lower 3bits to align the address to 8bytes.
+//   return (char*)(((size_t)ptr & (~(size_t)0x07)) + 0x08);
+// }
+// size_t align64ForTest(size_t ptr) {
+//   return (size_t)align64((char*)ptr);
+// }
 import "C"
 
 import (
@@ -61,10 +86,7 @@ func (s *shmImage) ColorModel() color.Model {
 }
 
 func (s *shmImage) Bounds() image.Rectangle {
-	return image.Rectangle{
-		Min: image.Point{X: 0, Y: 0},
-		Max: image.Point{X: int(s.img.width), Y: int(s.img.height)},
-	}
+	return image.Rect(0, 0, int(s.img.width), int(s.img.height))
 }
 
 type colorFunc func() (r, g, b, a uint32)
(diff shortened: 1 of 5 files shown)


pkg/io/video/convert.go
```diff
@@ -28,27 +28,44 @@ func imageToYCbCr(dst *image.YCbCr, src image.Image) {
...
-		b := make([]uint8, i2)
-		dst.Y = b[:i0:i0]
-		dst.Cb = b[i0:i1:i1]
-		dst.Cr = b[i1:i2:i2]
+		if cap(dst.Y) < 3*flat {
+			dst.Y = make([]uint8, i2)
+		}
+		dst.Y = dst.Y[:i0]
+		dst.Cb = dst.Y[i0:i1]
+		dst.Cr = dst.Y[i1:i2]
```
at-wat: Keep capacity to avoid realloc on reusing.

pkg/io/video/convert.go
```diff
@@ -28,27 +28,44 @@ func imageToYCbCr(dst *image.YCbCr, src image.Image) {
 		i0 := 1 * flat
 		i1 := 2 * flat
 		i2 := 3 * flat
-		b := make([]uint8, i2)
-		dst.Y = b[:i0:i0]
-		dst.Cb = b[i0:i1:i1]
-		dst.Cr = b[i1:i2:i2]
+		if cap(dst.Y) < 3*flat {
+			dst.Y = make([]uint8, i2)
```
at-wat: Realloc only if it lacks capacity.

pkg/io/video/convert.go
```diff
@@ -28,27 +28,44 @@ func imageToYCbCr(dst *image.YCbCr, src image.Image) {
...
-			// TODO: probably try to get the alpha value with something like
-			// https://en.wikipedia.org/wiki/Alpha_compositing
-			r, g, b, _ := src.At(xi, yi).RGBA()
-			yy, cb, cr := color.RGBToYCbCr(uint8(r/256), uint8(g/256), uint8(b/256))
-			dst.Y[i] = yy
-			dst.Cb[i] = cb
-			dst.Cr[i] = cr
-			i++
+	switch s := src.(type) {
+	case *image.RGBA:
```
at-wat: Add RGBA specialized version.

at-wat:
After applying this and https://github.com/pion/mediadevices/pull/55/commits/c46a89b4ba5b12559bdbc8f0a9b997098ffb9d90, CPU usage of screen casting example decreases to 40% on my laptop, which was saturated to 100% before.
(but, still twice or three times heavyer than gstreamer)

lherman-cs:
Oops, sorry. I didn't see this one. Could change the target to be master instead.

at-wat:
rebeased

pkg/io/video/convert.go
lherman-cs: knit: could you make it `i2` instead of `3*flat`?

pkg/driver/screen/x11capture_linux.go
```diff
@@ -1,11 +1,31 @@
...
+//     d++;
+//     s++;
+//   }
+// }
+// char *align64(char *ptr) { // return 64bit aligned pointer
+//   if (((size_t)ptr & 0xFF) == 0) {
+//     return ptr;
+//   }
+//   return (char*)(((size_t)ptr & (~(size_t)0xFF)) + 0x100);
+// }
```
lherman-cs: I couldn't read this... ^^, do you mind sharing what this does?

at-wat: (oops, 0xFF should be 0x07, and 0x100 should be 0x08.)
It clears lower 3 bits by `& ~0x07` (on 32bit env, it's equivalent of `& 0xFFFFFFF8`) and add 8 if there was cleared bits to avoid entering lower bound of the memory region.
So, the address gets multiple of 8bytes.

pkg/driver/screen/x11capture_linux.go
```diff
@@ -93,7 +131,7 @@ func newShmImage(dp *C.Display, screen int) (*shmImage, error) {
 
 	s := &shmImage{dp: dp}
 
-	s.shm.shmid = C.shmget(C.IPC_PRIVATE, C.ulong(w*h*4), C.IPC_CREAT|0600)
+	s.shm.shmid = C.shmget(C.IPC_PRIVATE, C.ulong(w*h*4+16), C.IPC_CREAT|0600)
```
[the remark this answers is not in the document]
at-wat: https://github.com/pion/mediadevices/pull/62/files#diff-39f75afe40196f59bfea2d519d3a6e2cR23-R28 may add offset to start address up to 8 bytes. +8 should be enough at not.

pkg/driver/screen/x11capture_linux.go
lherman-cs: I still don't get it.... but, oh well, it's alright. thanks

at-wat: It's basically for speeding up memory read/write.
64bit CPU can read/write 64bit data by single operation if the address is aligned. (In most case of recent environment, compiler and core library automatically align it on allocation.)

PR #64, opened 2020-02-16
Add video throttle transform

### LINKED ISSUE #63: Add video throttle transform
Drop frames to limit output framerate.

at-wat:
### Reference issue
Fixes #63

### DIFF
--- a/pkg/io/video/throttle.go
+++ b/pkg/io/video/throttle.go
@@ -1,0 +1,28 @@
+package video
+
+import (
+	"image"
+	"time"
+)
+
+// Throttle returns video throttling transform.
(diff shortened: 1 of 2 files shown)


pkg/io/video/throttle.go
```diff
@@ -0,0 +1,27 @@
+package video
+
+import (
+	"image"
+	"time"
+)
+
+// Scale returns video throttling transform.
```
lherman-cs: knit: could you also add some comments about what rate is? I think, from my understanding, rate is frame per second. Am I right?

at-wat: That's right.
I fixed the comment.

Issue #61, opened 2020-02-16
Pass also requested properties to Video/AudioRecord


at-wat:
Only selected prop is passed to `Video/AudioRecord` at now.

For example on screen capture driver, FrameRate is not discrete.
It would be nice to pass the selected prop and also the requested prop to read such parameters.

lherman-cs:
I definitely forgot about handling non-discrete values in the design. At first, I figured that it was nice to have this design because then you can do a sanity check in the wrapper level, and let the adapter implementors to simply use the prop.

lherman-cs:
As you have suggested in #69, it would be nice to have an interface from the wrapper to get the default value. 

But, I think instead of putting the interface in Wrapper, we can put it under `prop` since the default values shouldn't be dependent on anything.

So, we probably have something like,

```go
func (p Media) Default() prop.Media
```
Or probably even have it in package level

```
package prop

var Default = prop.Media{
  ...
}
```

What do you think?

at-wat:
I think the default values (especially continuous parameters) depends on the actual device.
For example camera would have default framerate of 30fps in most case. But, screen capture maybe 10fps since it's too heavy at 30fps.

lherman-cs:
I see. If that's the case, I'm not seeing that we can do better than we already have rn, set the default value in each adapter implementation.

PR #71, opened 2020-02-18
Pass user requested prop to drivers

at-wat:
#61

### DIFF
--- a/pkg/driver/screen/x11_linux.go
+++ b/pkg/driver/screen/x11_linux.go
@@ -59,11 +59,11 @@ func (s *screen) Close() error {
 	return nil
 }
 
-func (s *screen) VideoRecord(p prop.Media) (video.Reader, error) {
-	if p.FrameRate == 0 {
-		p.FrameRate = 10
+func (s *screen) VideoRecord(p, req prop.Media) (video.Reader, error) {
+	if req.FrameRate == 0 {
+		req.FrameRate = 10
 	}
-	s.tick = time.NewTicker(time.Duration(float32(time.Second) / p.FrameRate))
+	s.tick = time.NewTicker(time.Duration(float32(time.Second) / req.FrameRate))
 
 	var dst image.RGBA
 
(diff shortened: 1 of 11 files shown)


pkg/driver/screen/x11_linux.go
```diff
@@ -59,11 +59,11 @@ func (s *screen) Close() error {
 	return nil
 }
 
-func (s *screen) VideoRecord(p prop.Media) (video.Reader, error) {
-	if p.FrameRate == 0 {
-		p.FrameRate = 10
+func (s *screen) VideoRecord(p, req prop.Media) (video.Reader, error) {
+	if req.FrameRate == 0 {
```
at-wat: `p` is a prop selected from the list returned by `Properties()`, so it can't contain user requested values for continuously settable parameters.

lherman-cs:
@at-wat Hey, I just created an alternative approach to solve this, https://github.com/pion/mediadevices/pull/73. Please let me know what you think.

lherman-cs:
I feel like this approach is fine. But, I'm afraid that it can cause confusion to the adapter implementors or the caller for driver (even more confusing for people who want to use it directly) by having an extra parameter with the same type while the two parameters complement each other.

at-wat:
https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints
> The **`MediaTrackConstraints`** dictionary is used to describe a set of capabilities and the value or values each can take on. A constraints dictionary is passed into **`applyConstraints()`** to allow a script to establish a set of exact (required) values or ranges and/or preferred values or ranges of values for the track, and the most recently-requested set of custom constraints can be retrieved by calling **`getConstraints()`**.

It might be better that `prop.Media` has unexposed copy of user requested `prop.Media` and its getter.
@lherman-cs How do you think?

at-wat:
Oops, `applyConstraints()` and `getConstraints()` are the method of `MediaStreamTrack`. It doesn't help this case.

at-wat:
fixed by #73

PR #72, opened 2020-02-18
Call OnEnded even if errored before register

at-wat:
OnEnded handler was not called if something errored before
registering the handler. Store error and call handler immediately
on handler registration.

### DIFF
--- a/track.go
+++ b/track.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"sync/atomic"
+	"sync"
 
 	"github.com/pion/mediadevices/pkg/codec"
 	"github.com/pion/mediadevices/pkg/driver"
@@ -19,6 +19,9 @@ type Tracker interface {
 	Track() *webrtc.Track
 	LocalTrack() LocalTrack
 	Stop()
+	// OnEnded registers a handler to receive an error from the media stream track.
+	// If the error is already occured before registering, the handler will be
+	// immediately called.
 	OnEnded(func(error))
 }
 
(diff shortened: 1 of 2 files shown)


track.go
```diff
@@ -66,13 +72,29 @@ func newTrack(codecs []*webrtc.RTPCodec, trackGenerator TrackGenerator, d driver
 }
 
 func (t *track) OnEnded(handler func(error)) {
-	t.onErrorHandler.Store(handler)
+	t.mu.Lock()
+	t.onErrorHandler = handler
+	err := t.err
+	t.mu.Unlock()
```
lherman-cs: probably use `defer` instead?

track.go
```diff
@@ -66,13 +72,29 @@ func newTrack(codecs []*webrtc.RTPCodec, trackGenerator TrackGenerator, d driver
...
+		})
+	}
 }
 
 func (t *track) onError(err error) {
-	handler := t.onErrorHandler.Load()
+	t.mu.Lock()
+	t.err = err
+	handler := t.onErrorHandler
+	t.mu.Unlock()
```
lherman-cs: same as above

track.go
```diff
@@ -66,13 +72,29 @@ func newTrack(codecs []*webrtc.RTPCodec, trackGenerator TrackGenerator, d driver
 }
 
 func (t *track) OnEnded(handler func(error)) {
-	t.onErrorHandler.Store(handler)
+	t.mu.Lock()
+	t.onErrorHandler = handler
+	err := t.err
+	t.mu.Unlock()
+
+	if err != nil {
```
lherman-cs: probably check if handler is nil, since caller can pass nil.

track.go
at-wat: I want to Unlock before calling `handler`, since user can call `OnEnded` in the handler which causes deadlock if not unlocked.

lherman-cs: Ah, I see. You really thought ahead.... Love it.

PR #73, opened 2020-02-19
Add support for non-discrete constraints

lherman-cs:
This PR contains an alternative approach to https://github.com/pion/mediadevices/pull/71.

Resolves https://github.com/pion/mediadevices/issues/61

* Add a Merge method in prop
* Replace a manual merge in mediadevices in best property selection with the new Merge method

### DIFF
--- a/pkg/prop/prop.go
+++ b/pkg/prop/prop.go
@@ -3,6 +3,7 @@ package prop
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/prop/prop.go
```diff
@@ -16,6 +17,35 @@ type Media struct {
...
+			fieldB := b.Field(i)
+
+			// if a is a struct, b is also a struct. Then,
+			// we recursively merge them
+			if fieldA.Kind() == reflect.Struct {
+				merge(fieldA, fieldB)
+				continue
+			}
+
+			if !fieldB.IsZero() {
```
at-wat: `IsZero` is available since 1.13.
```suggestion
			if !reflect.DeepEqual(reflect.Zero(fieldB), fieldB) {
```
would be a Go 1.12 compatible equivalent.

lherman-cs: Oops, I just saw your comment. This makes more readable. I'll update the PR

lherman-cs: I just tested it. It didn't seem to work... Some of the tests failed. `fieldB.Interface() != reflect.Zero(fieldB.Type()).Interface()` seems to be working better.

lherman-cs:
> It seems fine if we will never have a Prop field which its zero value has meaning.

Right. Are there any parameters that may use zero values?

pkg/prop/prop.go
at-wat: Oops, I didn't tested that.
Current one seems nice.

at-wat:
https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints
Boolean things may have problem.

lherman-cs:
hmm, that's true...

at-wat:
Another idea is making everything functional options:
```go
type Audio struct {
  ChannelCount *int
  Latency      *time.Duration
  SampleRate   *int
  SampleSize   *int
} 
  
type MediaOption func(*Media)
  
func AudioChannelCount(n int) MediaOption {
  return func(p *Media) {
    p.ChannelCount = &n
  }
}
func AudioLatency(d time.Duration) MediaOption {
  return func(p *Media) {
    p.AudioLatency = &d
  }
}

md.GetUserMedia(
  AudioChannelCount(2),
  AudioLatency(time.Millisecond),
  ...
)
```

lherman-cs:
From my understanding, we're still checking the zero values, but since they're pointers we will be fine with boolean because now we have 3 possibilities, true, false, and nil. Am I close?

at-wat:
> since they're pointers we will be fine with boolean because now we have 3 possibilities, true, false, and nil.

Yes. Instead, assignment of the field is troublesome.

Or, it might be easier to just define three state boolean type.

lherman-cs:
I actually have another idea. How about if we simply make an exception for boolean? Since the original idea was to handle non-discrete values anyway, and boolean is not.

at-wat:
It sounds reasonable to me.

lherman-cs:
Sweet. I'll make the change then.

lherman-cs:
@at-wat Updated the PR with boolean exception.

PR #75, opened 2020-02-19
Close drivers on stream initialization failure

### LINKED ISSUE #74: GetUserMedia doesn't close driver if request is partially successed
When calling GetUserMedia with audio and video constraints, if one of audio/video failed to be initialized, other one is left opened. Then, the next GetUserMedia call always fail with `invalid state: driver is already opened`.

at-wat:
In GetUserMedia and GetDisplayMedia, driver was left opened if
initialization failure. Since Trackers are not returned on error,
there was no way to close them.
This commit closes Trackers on GetUserMedia/GetDisplayMedia failure.

### Reference issue
Fixes #74

### DIFF
--- a/mediadevices.go
+++ b/mediadevices.go
@@ -82,6 +82,12 @@ func WithTrackGenerator(gen TrackGenerator) MediaDevicesOption {
 func (m *mediaDevices) GetDisplayMedia(constraints MediaStreamConstraints) (MediaStream, error) {
 	trackers := make([]Tracker, 0)
 
+	cleanTrackers := func() {
+		for _, t := range trackers {
+			t.Stop()
+		}
+	}
+
 	var videoConstraints MediaTrackConstraints
 	if constraints.Video != nil {
 		constraints.Video(&videoConstraints)
@@ -90,6 +96,7 @@ func (m *mediaDevices) GetDisplayMedia(constraints MediaStreamConstraints) (Medi
 	if videoConstraints.Enabled {
 		tracker, err := m.selectScreen(videoConstraints)
 		if err != nil {
+			cleanTrackers()
 			return nil, err
 		}
 
@@ -98,6 +105,7 @@ func (m *mediaDevices) GetDisplayMedia(constraints MediaStreamConstraints) (Medi
 
 	s, err := NewMediaStream(trackers...)
 	if err != nil {
+		cleanTrackers()
 		return nil, err
 	}
 
@@ -111,6 +119,12 @@ func (m *mediaDevices) GetUserMedia(constraints MediaStreamConstraints) (MediaSt
 	// TODO: It should return media stream based on constraints
 	trackers := make([]Tracker, 0)
 
+	cleanTrackers := func() {
+		for _, t := range trackers {
+			t.Stop()
+		}
+	}
+
 	var videoConstraints, audioConstraints MediaTrackConstraints
 	if constraints.Video != nil {
 		constraints.Video(&videoConstraints)
@@ -123,6 +137,7 @@ func (m *mediaDevices) GetUserMedia(constraints MediaStreamConstraints) (MediaSt
 	if videoConstraints.Enabled {
 		tracker, err := m.selectVideo(videoConstraints)
 		if err != nil {
+			cleanTrackers()
 			return nil, err
 		}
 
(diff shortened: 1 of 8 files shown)


mediadevices.go
```diff
@@ -82,6 +82,12 @@ func WithTrackGenerator(gen TrackGenerator) MediaDevicesOption {
 func (m *mediaDevices) GetDisplayMedia(constraints MediaStreamConstraints) (MediaStream, error) {
 	trackers := make([]Tracker, 0)
 
+	cleanTrackers := func() {
+		for _, t := range trackers {
+			t.Stop()
+		}
+	}
+
```
lherman-cs: I like it!

mediadevices_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/pion/mediadevices/pkg/io/audio"
+	"github.com/pion/mediadevices/pkg/io/video"
+	"github.com/pion/mediadevices/pkg/prop"
+)
+
+func TestGetUserMedia(t *testing.T) {
+	codec.Register("MockVideo", codec.VideoEncoderBuilder(func(r video.Reader, p prop.Media) (io.ReadCloser, error) {
+		if p.BitRate == 0 {
+			return nil, errors.New("wrong codec parameter")
+		}
```
lherman-cs: Why is this an error?

mediadevices_test.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	// GetUserMedia with correct parameters
+	ms, err = md.GetUserMedia(constraints)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	tracks := ms.GetTracks()
+	if l := len(tracks); l != 2 {
+		t.Fatalf("Number of the tracks is expected to be 1, got %d", l)
```
lherman-cs: Oops, this should be "expected to be 2"

mediadevices_test.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	tracks[0].OnEnded(func(err error) {
+		if err != io.EOF {
+			t.Errorf("OnEnded called: %v", err)
+		}
+	})
```
lherman-cs: Probably also check for the audio track?

mediadevices_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// Stop and retry GetUserMedia
+	ms, err = md.GetUserMedia(constraints)
+	if err != nil {
+		t.Fatalf("Failed to GetUserMedia after the previsous tracks stopped: %v", err)
+	}
+	tracks = ms.GetTracks()
+	if l := len(tracks); l != 2 {
+		t.Fatalf("Number of the tracks is expected to be 1, got %d", l)
+	}
```
lherman-cs: Should be "expected to be 2" also

mediadevices_test.go
```diff
@@ -0,0 +1,179 @@
...
+	tracks[0].OnEnded(func(err error) {
+		if err != io.EOF {
+			t.Errorf("OnEnded called: %v", err)
+		}
+	})
```
lherman-cs: Probably also check for the audio track

track.go
```diff
@@ -108,6 +108,7 @@ func newVideoTrack(opts *MediaDevicesOptions, d driver.Driver, constraints Media
 	vr := d.(driver.VideoRecorder)
 	r, err := vr.VideoRecord(constraints.Media)
 	if err != nil {
+		_ = d.Close()
```
lherman-cs: I think we should just Close the driver automatically in `driver` package. And, we should simply call the wrapper version of Close so the driver state will be updated too and finally still return the error. It looks to me that this is the `driver` package problem. What do you think?

mediadevices_test.go
at-wat: It's just for triggering error during initialization process for testing.

track.go
at-wat: Ah, yes. This can be done in the wrapper.

mediadevices_test.go
lherman-cs: Ah I see

track.go
at-wat: image
So, there will be a choice pseudo-state at the place of `Audio/VideoRecord` and an another one arrow to `Closed` state on failure case.

at-wat:
@lherman-cs addressed your comments!

track.go
lherman-cs: Makes sense. Thanks for pointing that out!

PR #81, opened 2020-02-22
Fix build without CGO

at-wat:
- Add build test without CGO
- Declare functions selected at runtime for non-CGO environment


at-wat:
rebasing

PR #83, opened 2020-02-22
Add microphone driver for Windows

at-wat:
Using classic waveInOpen API.

Some of the part is stub, but it works for default input device.
Future works: enum devices, detect and handle microphone device disconnection, get actual props

fixes #83

### DIFF
--- a/pkg/driver/microphone/microphone_windows.go
+++ b/pkg/driver/microphone/microphone_windows.go
@@ -12,6 +12,12 @@ import (
 	"github.com/pion/mediadevices/pkg/prop"
 )
 
+const (
+	// bufferNumber * prop.Audio.Latency is the maximum blockable duration
+	// to get data without dropping chunks.
+	bufferNumber = 5
+)
+
 // Windows APIs
 var (
 	winmm                 = windows.NewLazySystemDLL("Winmm.dll")
@@ -34,6 +40,9 @@ func newBuffer(samples int) *buffer {
 	b := make([]int16, samples)
 	return &buffer{
 		waveHdr: waveHdr{
+			// Sharing Go memory with Windows C API without reference.
+			// Make sure that the lifetime of the buffer struct is longer
+			// than the final access from cbWaveIn.
 			lpData:         uintptr(unsafe.Pointer(&b[0])),
 			dwBufferLength: uint32(samples * 2),
 		},
@@ -88,6 +97,7 @@ func (m *microphone) Close() error {
 	if err := errWinmm[ret]; err != nil {
 		return err
 	}
+	// All enqueued buffers are marked done by waveInReset.
 	ret, _, _ = waveInReset.Call(
 		uintptr(unsafe.Pointer(m.hWaveIn)),
 	)
@@ -95,6 +105,7 @@ func (m *microphone) Close() error {
 		return err
 	}
 	for _, buf := range m.buf {
+		// Detach buffers from waveIn API.
 		ret, _, _ := waveInUnprepareHeader.Call(
 			uintptr(unsafe.Pointer(m.hWaveIn)),
 			uintptr(unsafe.Pointer(&buf.waveHdr)),
@@ -104,6 +115,10 @@ func (m *microphone) Close() error {
 			return err
 		}
 	}
+	// Now, it's ready to free the buffers.
+	// As microphone struct still has reference to the buffers,
+	// they will be GC-ed once microphone is reopened or unreferenced.
+
 	ret, _, _ = waveInClose.Call(
 		uintptr(unsafe.Pointer(m.hWaveIn)),
 	)
@@ -117,10 +132,12 @@ func (m *microphone) Close() error {
 }
 
 func (m *microphone) AudioRecord(p prop.Media) (audio.Reader, error) {
-	for i := 0; i < 5; i++ {
+	for i := 0; i < bufferNumber; i++ {
 		b := newBuffer(
 			int(uint64(p.Latency) * uint64(p.SampleRate) / uint64(time.Second)),
 		)
+		// Map the buffer by its data head address to restore access to the Go struct
+		// in callback function. Don't resize the buffer after it.
 		m.buf[uintptr(unsafe.Pointer(&b.waveHdr))] = b
 	}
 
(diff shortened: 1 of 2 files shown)


at-wat:
In my environment, it can be built by:
```
GOOS=windows
GOARCH=amd64
CGO_ENABLED=1
CC=x86_64-w64-mingw32-gcc
PKG_CONFIG=x86_64-w64-mingw32-pkg-config
go build .
```
(openh264 driver should be un-imported and libopus/libopusfile/libvpx for mingw are required)

Looks working on wine and Windows 10.

lherman-cs:
I've replaced openh264 with x264 instead, https://github.com/pion/mediadevices/pull/84.

lherman-cs:
Sorry for the delay. I never compiled codes in Windows... I'm not sure how to do it. So, you mentioned that you could use wine to run the binary. Is there a way to use wine to compile it too? It seems to be easier than having a Windows VM.

at-wat:
I compiled this on Linux using mingw64. `Wine` automatically wraps Windows APIs when executing .exe on the shell.
On Ubuntu, the toolchain can be installed as `gcc-mingw-w64, g++-mingw-w64` by apt. On Fedora, it's `mingw64-gcc, mingw64-g++`.

lherman-cs:
How does it compile to exe with the dynamic libraries? Aren't the dynamic libraries already precompiled for Linux?

at-wat:
On Fedora, mingw64 libopus/opusfile are provided as binary packages. I needed to build vpx from source.
Sadly, Ubuntu seems not having them.

lherman-cs:
I'm working on Dockerfiles for development here, https://github.com/pion/mediadevices/pull/91. The goal is to make testing easier with Linux.

lherman-cs:
I've successfully compiled it, but when I tried to run it, I got:

```
root@d67e4b7cbed9:/go/src/github.com/pion/mediadevices/examples/simple# file simple.exe 
simple.exe: PE32+ executable (console) x86-64, for MS Windows
root@d67e4b7cbed9:/go/src/github.com/pion/mediadevices/examples/simple# wine simple.exe 
wine: Bad EXE format for Z:\go\src\github.com\pion\mediadevices\examples\simple\simple.exe.
root@d67e4b7cbed9:/go/src/github.com/pion/mediadevices/examples/simple# 
```

lherman-cs:
So, I found that the wine version that I installed was broken. After I fixed that, I got:

```
root@5fe9a5ad6ac2:/go/src/github.com/pion/mediadevices/examples/simple# wine simple.exe 
002a:err:module:import_dll Library libopus-0.dll (which is needed by L"Z:\\go\\src\\github.com\\pion\\mediadevices\\examples\\simple\\simple.exe") not found
002a:err:module:LdrInitializeThunk Importing dlls for L"Z:\\go\\src\\github.com\\pion\\mediadevices\\examples\\simple\\simple.exe" failed, status c0000135
```

My pkg-config (I compiled opus with mingw):
```
root@5fe9a5ad6ac2:/go/src/github.com/pion/mediadevices/examples/simple# pkg-config --list-all
opus             Opus - Opus IETF audio codec (floating-point build)
xkeyboard-config XKeyboardConfig - X Keyboard configuration data
udev             udev - udev
iso-codes        iso-codes - ISO country, language, script and currency codes and translations
shared-mime-info shared-mime-info - Freedesktop common MIME database
systemd          systemd - systemd System and Service Manager
```

lherman-cs:
So, somehow, libopus put its dynamic library, `libopus-0.dll` in `/usr/local/bin` which is not the in the default `LD_LIBRARY_PATH`. But, I fixed that and now I'm getting,

```
root@5fe9a5ad6ac2:/usr/local/bin# echo $SDP | wine /go/src/github.com/pion/mediadevices/examples/simple/simple.exe
002b:fixme:process:SetProcessPriorityBoost (FFFFFFFFFFFFFFFF,1): stub

002b:fixme:winsock:WSAIoctl WS_SIO_UDP_CONNRESET stub
002b:fixme:winsock:WSAIoctl WS_SIO_UDP_CONNRESET stub
002b:fixme:winsock:WSAIoctl WS_SIO_UDP_CONNRESET stub
002e:fixme:winsock:WSAIoctl WS_SIO_UDP_CONNRESET stub
ALSA lib pcm_dmix.c:1108:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm_dsnoop.c:638:(snd_pcm_dsnoop_open) unable to open slave
panic: bad device id

goroutine 1 [running]:
main.main()
        /go/src/github.com/pion/mediadevices/examples/simple/main.go:66 +0x74a
```

Note: I shared the microphone device by specifying `--device /dev/snd:/dev/snd` in the docker command.

lherman-cs:
I'm not sure what happened. But, I just reran the docker container, and it worked fine. I could hear my voice!!

at-wat:
I think features of this and #89 is ready as initial implementations.
I don't usually develop softwares for Windows, so I hope someone actually using Windows improves them.

pkg/driver/microphone/microphone_windows.go
```diff
@@ -0,0 +1,329 @@
...
+	return nil
+}
+
+func (m *microphone) AudioRecord(p prop.Media) (audio.Reader, error) {
+	for i := 0; i < 5; i++ {
+		b := newBuffer(
+			int(uint64(p.Latency) * uint64(p.SampleRate) / uint64(time.Second)),
+		)
+		m.buf[uintptr(unsafe.Pointer(&b.waveHdr))] = b
+	}
```
lherman-cs: I'm not sure what this for loop is being used for. Why are we creating the same buffer 5 times?

at-wat: This API needs multiple buffering (at least double buffer) to get signal continuously.

lherman-cs: hmm, but, why 5 I guess?

at-wat: It preserves received buffer until `audio.Reader.Read` is called.
5 has no specific reason, but it means that the maximum blockable duration to get data without dropping chunks is 5 * `p.Latency` = 100ms at now. `Read` timing jitter or delay larger than this will make the sound choppy.

lherman-cs: Ah, I see, do you mind to add this as a comment?

at-wat: I'll make it named constant and add a comment.

lherman-cs: thanks!

at-wat: Added some more comments especially about buffering.

PR #84, opened 2020-02-22
Add X264 encoder

lherman-cs:
**Description**
* Add x264 codec (requires `libvpx-dev` to be installed)
* Support KeyFrameInterval
* Support Quality rate control

### DIFF
--- a/pkg/codec/x264/bridge.h
+++ b/pkg/codec/x264/bridge.h
@@ -1,0 +1,88 @@
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <x264.h>
+
+typedef struct Slice {
+  unsigned char *data;
+  int data_len;
+} Slice;
+
+typedef struct Encoder {
+  x264_t *h;
+  x264_picture_t pic_in, pic_out;
+  x264_param_t param;
+} Encoder;
+
+Encoder *enc_new(x264_param_t param) {
+  Encoder *e = (Encoder *)malloc(sizeof(Encoder));
+
+  if (x264_param_default_preset(&e->param, "veryfast", "zerolatency") < 0)
+    goto fail;
+
+  /* Configure non-default params */
+  e->param.i_csp = param.i_csp;
+  e->param.i_width = param.i_width;
+  e->param.i_height = param.i_height;
+  e->param.i_fps_num = param.i_fps_num;
+  e->param.i_fps_den = 1;
+  // Intra refres:
+  e->param.i_keyint_max = param.i_fps_num;
+  e->param.b_intra_refresh = 1;
+  // Rate control:
+  e->param.rc.i_rc_method = X264_RC_CRF;
+  e->param.rc.f_rf_constant = 25;
+  e->param.rc.f_rf_constant_max = 35;
+  // For streaming:
+  e->param.b_repeat_headers = 1;
+  e->param.b_annexb = 1;
+
+  if (x264_param_apply_profile(&e->param, "baseline") < 0)
+    goto fail;
+
+  if (x264_picture_alloc(&e->pic_in, param.i_csp, param.i_width, param.i_height) < 0)
+    goto fail;
+
+#undef fail
+#define fail fail2
+  e->h = x264_encoder_open(&e->param);
+  if (!e->h)
+    goto fail2;
+
+  return e;
+
+#undef fail
+fail2:
+  x264_picture_clean(&e->pic_in);
+
+fail:
+  free(e);
+  // TODO: set appropriate errno
+  errno = -1;
+  return NULL;
+}
+
+Slice enc_encode(Encoder *e, uint8_t *y, uint8_t *cb, uint8_t *cr) {
+  x264_nal_t *nal;
+  int i_nal;
+
+  e->pic_in.img.plane[0] = y;
+  e->pic_in.img.plane[1] = cb;
+  e->pic_in.img.plane[2] = cr;
+
+  int frame_size = x264_encoder_encode(e->h, &nal, &i_nal, &e->pic_in, &e->pic_out);
+  Slice s = {.data_len = frame_size};
+  if (frame_size <= 0) {
+    return s;
+  }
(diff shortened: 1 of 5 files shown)


pkg/codec/x264/bridge.h
```diff
@@ -0,0 +1,88 @@
...
+  e->param.b_annexb = 1;
+
+  if (x264_param_apply_profile(&e->param, "baseline") < 0)
+    goto fail;
+
+  if (x264_picture_alloc(&e->pic_in, param.i_csp, param.i_width, param.i_height) < 0)
+    goto fail;
+
+#undef fail
+#define fail fail2
```
at-wat: `fail` label is not used later. Just `undef` seems enough to avoid leaking.

pkg/codec/x264/bridge.h
```diff
@@ -0,0 +1,88 @@
...
+    goto fail;
+
+  /* Configure non-default params */
+  e->param.i_csp = param.i_csp;
+  e->param.i_width = param.i_width;
+  e->param.i_height = param.i_height;
+  e->param.i_fps_num = param.i_fps_num;
+  e->param.i_fps_den = 1;
+  // Intra refres:
+  e->param.i_keyint_max = param.i_fps_num;
```
at-wat: https://github.com/pion/mediadevices/blob/6fd00ab14568e16971363b1622ba58c9feb9b15b/pkg/prop/prop.go#L128
`Codec.KeyFrameInterval` can be used.

pkg/codec/x264/bridge.h
lherman-cs: I've removed it, instead uses fail2 directly instead.

pkg/codec/x264/bridge.h
lherman-cs: Oops, makes sense. Updated it.

lherman-cs:
@at-wat I've updated the PR. PTAL.

lherman-cs:
It seems like we need a better way of controlling the rate control. For example, `Bitrate` and `Quality` seem to be mutually exclusive, `Quality` impacts `Bitrate` and also the other way around.

at-wat:
> For example, Bitrate and Quality seem to be mutually exclusive, Quality impacts Bitrate and also the other way around.

In codecs, quality is separately configurable. It's usually a complexity of inter frame prediction. By using more CPU power to it, same size of the compressed data can represents larger amount of information.

lherman-cs:
Wait, so am I interpreting the RF parameter wrong? I'm using the Quality to control RF.

at-wat:
I'm not very sure, but in 
https://github.com/mirror/x264/blob/1771b556ee45207f8711744ccbd5d42a3949b14c/x264.h#L441-L442
`f_rf_constant` seems like a VBR control parameter and `i_bitrate` also exists.

lherman-cs:
So, apparently, if we choose to use CRF mode, `x264` will reset `i_bitrate` to 0 (ignore), https://github.com/mirror/x264/blob/1771b556ee45207f8711744ccbd5d42a3949b14c/encoder/encoder.c#L874.

pkg/codec/x264/bridge.h
```diff
@@ -0,0 +1,94 @@
...
+  /* Configure non-default params */
+  e->param.i_csp = param.i_csp;
+  e->param.i_width = param.i_width;
+  e->param.i_height = param.i_height;
+  e->param.i_fps_num = param.i_fps_num;
+  e->param.i_fps_den = 1;
+  // Intra refres:
+  e->param.i_keyint_max = param.i_keyint_max;
+  // Rate control:
+  e->param.rc.i_rc_method = X264_RC_CRF;
```
at-wat: I think VBR mode is better especially for WebRTC.
Bitrate of CRF depends on the input video and takes high bitrate if entropy of the input is large.

at-wat:
Encoding complexity can be configured like https://github.com/mirror/x264/blob/1771b556ee45207f8711744ccbd5d42a3949b14c/common/base.c#L434-L540.
Slower encoding let output quality higher for same bitrate.

lherman-cs:
According to this article https://slhck.info/video/2017/03/01/rate-control.html, both ABR and CRF seem to be fine rate controls. But, what I'm missing here is VBV-constained bitrate to make sure that we're encoding under the bandwidth limit.

lherman-cs:
Looking at this block of code https://github.com/mirror/x264/blob/master/encoder/encoder.c#L1838-L1848, it seems that we only need to set `i_vbv_max_bitrate` and `i_vbv_buffer_size` to be greater than zero

lherman-cs:
> According to this article https://slhck.info/video/2017/03/01/rate-control.html, both ABR and CRF seem to be fine rate controls. But, what I'm missing here is VBV-constained bitrate to make sure that we're encoding under the bandwidth limit.

After a second thought, ABR seems to be more suitable for WebRTC like you said. As the goal in WebRTC is usually to meet a bitrate target, we can use ABR to control it based on the average bitrate to meet this bitrate target from the user.

Then, finally, we can get some help from VBV constraint to meet the target bitrate by using a buffer.

lherman-cs:
@at-wat I just updated the PR to use ABR+VBV. PTAL.

pkg/codec/x264/x264.go
```diff
@@ -0,0 +1,159 @@
...
+		i_width:      C.int(p.Width),
+		i_height:     C.int(p.Height),
+		i_keyint_max: C.int(p.KeyFrameInterval),
+	}
+	param.rc.i_bitrate = C.int(p.BitRate)
+	param.rc.i_vbv_max_bitrate = param.rc.i_bitrate
+	param.rc.i_vbv_buffer_size = param.rc.i_vbv_max_bitrate * 2
+
+	var rc C.int
+	cPreset := C.CString(fmt.Sprint(preset))
```
at-wat: maybe nice to have a comment that `cPreset` is freed in `enc_new`.

lherman-cs: ah, yep

Issue #82, opened 2020-02-22
OpenH264 link on non-x86_64-Linux environment


at-wat:
`cvendor/lib/openh264/libopenh264.a` is compiled for x86_64 Linux.
Would be nice to have more portable way for multiple environments.

lherman-cs:
This has been my goal. The original thought was to bundle everything statically and reduce the number of dependencies that you need to install (basically I want people to be able to just get the library by `go get` and just build their code with it, no need `apt-get install` or anything like that). My aggressive goal is to actually write these codecs in pure Go, but obviously we don't have the time for that yet... ^^

From my research in the past, we can build the libraries statically for every OS and architecture, and link them conditionally in Go (which is somewhat what we currently have, we just need to extend more support to other OS and arch).

In fact, this approach has been done before in a project called, [mysterium](https://medium.com/mysterium-network/golang-c-interoperability-caf0ba9f7bf3):
* Blog post specifically about this, https://medium.com/mysterium-network/golang-c-interoperability-caf0ba9f7bf3
* The source codes, https://github.com/mysteriumnetwork/go-openvpn/tree/master/openvpn3

Honestly, the main problem has been getting hardware with different OSes and arch to compile it statically. I saw some articles that you could use Docker to get the proper toolchain for cross-compilation. But, I'm not sure how we can do cross-OS compilations since Windows and Mac toolchain can't be used in Linux AFAIK.

Also, it seems that making a static build with cgo might become easier in the future with this proposal, https://github.com/golang/go/issues/26492. I really wonder that if it's going to be possible to cross-compile with cgo in the future... It would be pretty awesome!

at-wat:
I agree that bundling everything statically is useful.
However, static link is sometimes problematic due to their licenses (currently, linked libs excepting libgcc are all BSD-like, so it's OK) and binary size when building multiple executables. It would be better to be selectable via `-tags` or something.

lherman-cs:
That makes sense. But, it seems like OpenH264 doesn't distribute their dynamic version to package managers... I'm having a hard time to find one for Ubuntu.

at-wat:
Hmm, it seems that Cisco's AVC/H.264 Patent Portfolio is only applied to their precompiled binary.

http://www.openh264.org/faq.html
http://www.openh264.org/BINARY_LICENSE.txt
[...]

PR #85, opened 2020-02-23
Add openh264 library for darwin and windows

at-wat:
Use dockercore/golang-cross to cross compile openh264.
Set `-tags dynamic` to use system installed library.

**~Windows/Darwin binary is not tested yet.~**

### Reference issue
Partially address #82 
Non-x86_64 arch is still missing.


at-wat:
Added a workaround to build with both MinGW MSYS 5.x and 6.x.
I confirmed that openh264 codec works on Windows 10.

at-wat:
Hope someone could test darwin version on MacOS.

lherman-cs:
Hey, I just tested it on mac, and I got this:

```
ld: warning: ignoring file ../../pkg/codec/openh264/../../../cvendor/lib/openh264/libopenh264.x86_64-darwin.a, file was built for archive which is not the architecture being linked (x86_64): ../../pkg/codec/openh264/../../../cvendor/lib/openh264/libopenh264.x86_64-darwin.a
Undefined symbols for architecture x86_64:
  "_WelsCreateSVCEncoder", referenced from:
      _enc_new in _x004.o
  "_WelsDestroySVCEncoder", referenced from:
      _enc_free in _x004.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It looks like that the problem is caused by the linker conflict since mac uses XCode by default instead of GNU. I'm double checking it with GNU toolchain instead.

Also, I found that the struct initialization that I used in `bridge.cpp` is not compatible with Clang. So, I made a small fix here,  https://github.com/pion/mediadevices/pull/86.

lherman-cs:
It doesn't seem that GNU `ld` supports Darwin sadly:
* https://www.quora.com/How-do-I-get-GNU-ld-on-OS-X
* https://github.com/Homebrew/homebrew-core/issues/17794
* https://apple.stackexchange.com/questions/341105/ld-linker-is-missing-when-installing-gcc-using-homebrew-on-mac-os

Makefile
```diff
@@ -3,16 +3,53 @@ src_dir = $(vendor_dir)/src
...
 
-vendor: create_out_dirs openh264
+make_args.x86_64-windows = \
+	CC=x86_64-w64-mingw32-gcc \
+	CXX=x86_64-w64-mingw32-g++ \
+	ARCH=x86_64 \
+	OS=mingw_nt
+make_args.x86_64-darwin = \
+	CC=o64-clang \
+	CXX=o64-clang++ \
```
at-wat: maybe `AR=llvm-ar` is required

at-wat: This is it. Cross compiling the example is fixed by this.

at-wat:
@lherman-cs I think the updated .a file should work for Darwin.

Makefile
lherman-cs: Yep, it works!

lherman-cs:
Btw, I just found another small bug... Somehow, the errno in `enc_new` was non-zero when it returned. I created another fix here, https://github.com/pion/mediadevices/pull/87.

cvendor/builder.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM dockercore/golang-cross
```
lherman-cs: I didn't know that this existed... TIL

Makefile
```diff
@@ -3,16 +3,54 @@ src_dir = $(vendor_dir)/src
...
+	docker build -t mediadevices-builder -f cvendor/builder.Dockerfile .
+	docker run -i --rm \
+		-v $(CURDIR):/go/src/github.com/pion/mediadevices \
+		mediadevices-builder make $(lib_dir)/openh264/libopenh264.x86_64-linux.a
+	docker run -i --rm \
+		-v $(CURDIR):/go/src/github.com/pion/mediadevices \
+		mediadevices-builder make $(lib_dir)/openh264/libopenh264.x86_64-windows.a
+	docker run -i --rm \
+		-v $(CURDIR):/go/src/github.com/pion/mediadevices \
+		mediadevices-builder make $(lib_dir)/openh264/libopenh264.x86_64-darwin.a
```
lherman-cs: `-i` doesn't seem necessary.

at-wat:
Removed unnecessary docker build option and squashed archive fix into the first commit.

PR #87, opened 2020-02-23
Fix incorrect errno

lherman-cs:


at-wat:
I think clearing `errno` is a caller's responsibility. Manpage says:

http://man7.org/linux/man-pages/man3/errno.3.html
> The value of errno is never set to zero by any system call or library function.

lherman-cs:
Hmm, that's interesting. This becomes really strange when we have to set `errno` from Go instead. What do you think?

at-wat:
I'm guessing that something is actually failing (or warning) inside the cpp functions.
Clearing errno at the beginning of the cpp functions which is called from Go would be safer.

lherman-cs:
I see. That makes sense actually. I'll update the PR

pkg/codec/openh264/bridge.cpp
```diff
@@ -51,8 +54,10 @@ Encoder *enc_new(const EncoderOptions opts) {
     errno = rv;
     return NULL;
   }
+  printf("3: %d\n", errno);
```
lherman-cs: So, somehow the errno got set to 60 (Operation timeout) here. But, when I tried to ignore this errno, it worked just fine.

at-wat: I found that errno is not goroutine safe on Darwin.
```go
package main
  
import (
  "fmt"
  "net"
  "time"
)

// #include <unistd.h>
// #include <errno.h>
// int wait() { usleep(2000000); }
import "C"

func main() {
  for i := 0; i < 10; i++ {
    go func() {
      for {
        _, _ = net.Dial("tcp", "localhost:12345")
        time.Sleep(time.Millisecond)
      }
    }()
  }
  _, err := C.wait()
  fmt.Printf("err: %v\n", err)
}
```
This code outputs `nil` on Linux, but on Darwin (tested on Travis-CI's OSX environment) outputs `operation timed out`.

at-wat: So, it seems unsafe to use errno on Darwin.
How about https://github.com/pion/mediadevices/pull/88?

lherman-cs: I always thought that the cgo error handling is pretty neat, using the second value for error by implying from errno. But, nevermind, it's not safe.... Oh, well, I like the approach, I don't know if there's a cleaner way of doing it atm.

lherman-cs:
Closing this in favor of https://github.com/pion/mediadevices/pull/88

PR #91, opened 2020-02-23
Add Dockerfiles for windows microphone development

lherman-cs:
#### Description
This PR enables Windows microphone development in Linux. This is a part of our bigger goal here, https://github.com/pion/mediadevices/issues/94.


windows.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ENV TARGET_OS=windows \
+    PYTHONUNBUFFERED=1 \
+    # Somehow libopus.dll is installed under /usr/local/bin, so we need to add this
+    # path to WINEPATH so that wine can find this library
+    WINEPATH=/usr/local/bin \
+    # Go and C default environments
+    GOOS=windows \
+    GOARCH=amd64 \
+    CGO_ENABLED=1 \
+    CC=x86_64-w64-mingw32-gcc
```
at-wat: CXX will also be needed to use OpenH264 and windows camera driver.

at-wat:
For Windows development, it might be easier to use Fedora as a base image apart from libraries builder.
Fedora has latest Wine and several libraries for MinGW in main repository.

windows.Dockerfile
lherman-cs: Ah, that's true. I forgot about that.

lherman-cs:
> For Windows development, it might be easier to use Fedora as a base image apart from libraries builder.
> Fedora has latest Wine and several libraries for MinGW in main repository.

So, I guess, we'll just have 4 images:
* builder.Dockerfile (I'll revert the change to put it back to cvendor)
* base.Dockerfile: I'll use fedora instead
* windows.Dockerfile
* darwin.Dockerfile

at-wat:
> - builder.Dockerfile
> - windows.Dockerfile: use fedora as base image instead
> - darwin.Dockerfile: use builder.Dockerfile as base image

I'm not sure about Darling environment, but it sounds good.

lherman-cs:
Honestly, I haven't tried it either. But, it looks like as long as it's not GUI, Darling can handle it.

windows.Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    CGO_ENABLED=1 \
+    GO111MODULE=on \
+    CC=x86_64-w64-mingw32-gcc \
+    CXX=x86_64-w64-mingw32-g++
+
+COPY scripts /usr/bin/ 
+
+RUN dnf install -y python3 diffutils make pkg-config mingw64-gcc mingw64-gcc-c++ nasm wine golang && \
+    dnf clean all && \
+    install-dependencies && \
```
at-wat: libopus for mingw64 is available as `mingw64-opus` in dnf.

lherman-cs: ouch...

lherman-cs: Thanks

lherman-cs:
I think I'll stop here for this PR. I think this is good enough to enable Windows microphone development. I'll continue working on this part later in different PRs.

Makefile
```diff
@@ -44,7 +44,7 @@ $(lib_dir)/openh264/libopenh264.x86_64-darwin.a: $(src_dir)/openh264
 
 .PHONY: cross-libraries
 cross-libraries:
-	docker build -t mediadevices-builder -f cvendor/builder.Dockerfile .
+	docker build -t mediadevices-builder -f builder.Dockerfile .
```
at-wat: Something like `libs-builder.Dockerfile` might be better?

windows.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+        python3 \
+        diffutils \
+        make \
+        pkg-config \
+        mingw64-gcc \
+        mingw64-gcc-c++ \
+        nasm \
+        wine \
+        golang \
+        mingw64-opus && \
```
at-wat: Would be better to be sorted.

Makefile
lherman-cs: sounds good. I'll rename it.

windows.Dockerfile
lherman-cs: Ah, makes sense

lherman-cs:
@at-wat PTAL. I've addressed your comments.

PR #97, opened 2020-02-25
Add box sampling scaler

at-wat:
NearestNeighbor | BoxSampling
--- | ---
Screenshot from 2020-02-26 00-15-54 | Screenshot from 2020-02-26 00-17-07

```
$ CGO_CFLAGS="-march=native -O3" go test . -bench BenchmarkScale/Fast -benchmem -benchtime 3s
goos: linux
goarch: amd64
pkg: github.com/pion/mediadevices/pkg/io/video
BenchmarkScale/FastNearestNeighbor/480p/RGBA-4         	    1419	   2156141 ns/op	     716 B/op	       1 allocs/op
BenchmarkScale/FastNearestNeighbor/480p/I444-4         	     441	   7967492 ns/op	    3566 B/op	       1 allocs/op
BenchmarkScale/FastNearestNeighbor/1080p/RGBA-4        	    1574	   2248581 ns/op	     652 B/op	       1 allocs/op
BenchmarkScale/FastNearestNeighbor/1080p/I444-4        	     426	   8358253 ns/op	    3687 B/op	       1 allocs/op
BenchmarkScale/FastBoxSampling/480p/RGBA-4             	     559	   6228782 ns/op	    8317 B/op	       1 allocs/op
BenchmarkScale/FastBoxSampling/480p/I444-4             	     416	   9021733 ns/op	    6001 B/op	       1 allocs/op
BenchmarkScale/FastBoxSampling/1080p/RGBA-4            	     144	  24268477 ns/op	   32102 B/op	       1 allocs/op
BenchmarkScale/FastBoxSampling/1080p/I444-4            	     100	  32458101 ns/op	   24561 B/op	       1 allocs/op
PASS
ok  	github.com/pion/mediadevices/pkg/io/video	37.338s
```

### DIFF
--- a/pkg/io/video/scaler_cgo.go
+++ b/pkg/io/video/scaler_cgo.go
@@ -5,13 +5,21 @@ package video
 import (
 	"golang.org/x/image/draw"
 	"image"
+	"sync"
 )
 
+// #include <stdint.h>
 // void fastNearestNeighbor(
-//     unsigned char* dst, unsigned const char* src,
+//     uint8_t* dst, const uint8_t* src,
 //     const int ch,
 //     const int dw, const int dh, const int dstride,
 //     const int sw, const int sh, const int sstride);
+// void fastBoxSampling(
+//     uint8_t* dst, const uint8_t* src,
+//     const int ch,
+//     const int dw, const int dh, const int dstride,
+//     const int sw, const int sh, const int sstride,
+//     uint32_t* tmp);
 import "C"
 
 // List of scaling algorithms
@@ -19,7 +27,16 @@ var (
 	// ScalerFastNearestNeighbor is a CGO version of NearestNeighbor scaler.
 	// This is roughly 4-times faster than draw.NearestNeighbor.
 	ScalerFastNearestNeighbor = Scaler(&FastNearestNeighbor{})
+	// ScalerFastBoxSampling is a CGO implementation of BoxSampling scaler.
+	// This is heavyer than NearestNeighbor but keeps detail on down scaling.
+	ScalerFastBoxSampling = Scaler(&FastBoxSampling{})
 )
+
+var poolBoxSampleBuffer = sync.Pool{
+	New: func() interface{} {
+		return &([]uint32{})
+	},
+}
 
 var (
 	scalerTestAlgosCGO = map[string]Scaler{
@@ -27,6 +44,7 @@ var (
 	}
 	scalerBenchAlgosCGO = map[string]Scaler{
 		"FastNearestNeighbor": ScalerFastNearestNeighbor,
+		"FastBoxSampling":     ScalerFastBoxSampling,
 	}
 )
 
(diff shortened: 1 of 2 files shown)


lherman-cs:
So, it looks like NearestNeighbour is faster than BoxSampling. I'm curious what pros does BoxSampling have?

pkg/io/video/scaler_cgo.go
```diff
@@ -92,3 +110,74 @@ func (f *FastNearestNeighbor) Scale(dst draw.Image, dr image.Rectangle, src imag
...
+		l := d.Stride * d.Rect.Dy()
+		if len(d.Pix) < l {
+			if cap(d.Pix) < l {
+				d.Pix = make([]uint8, l)
+			}
+			d.Pix = d.Pix[:l]
+		}
+		if cap(*tmp) < l {
+			*tmp = make([]uint32, l)
+		}
```
lherman-cs: I think `len(*tmp)` makes more sense here. I feel like it's a little bit weird if let say, we have a len of 10 and a cap of 20. Since C has access to the whole memory, 0 to 19, C can technically just write to it. And, from Go, when we append the slice, the new memory would be non-zero.

pkg/io/video/scaler_cgo.go
```diff
@@ -92,3 +110,74 @@ func (f *FastNearestNeighbor) Scale(dst draw.Image, dr image.Rectangle, src imag
...
+func (f *FastBoxSampling) Scale(dst draw.Image, dr image.Rectangle, src image.Image, sr image.Rectangle, op draw.Op, opts *draw.Options) {
+	ddx, ddy := dr.Dx(), dr.Dy()
+	sdx, sdy := sr.Dx(), sr.Dy()
+	if sdx < ddx && sdy < ddy {
+		// Box upsampling is equivalent of NearestNeighbor
+		(&FastNearestNeighbor{}).Scale(dst, dr, src, sr, op, opts)
+		return
+	}
+	tmp := poolBoxSampleBuffer.Get().(*[]uint32)
+	defer poolBoxSampleBuffer.Put(tmp)
```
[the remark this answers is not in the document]
at-wat: We can have two `MediaStream`s to concurrently use two cameras, so the buffer can't be a global buffer.
Also, since `Scaler` instance is shared in the process as it is defined like https://github.com/golang/image/blob/58c23975cae11f062d4b3b0c143fe248faac195d/draw/scale.go#L154-L158, it can't be a member of scaler struct.

pkg/io/video/scaler_cgo.go
at-wat: I had intended to avoid reallocation when length is smaller but capacity is enough. However, this temporary buffer is only used by this function and length is never trimmed, so `len()` would be fine.

pkg/io/video/scaler_cgo.go
```diff
@@ -92,3 +110,74 @@ func (f *FastNearestNeighbor) Scale(dst draw.Image, dr image.Rectangle, src imag
...
+
+// FastBoxSampling is a CGO version of Box sampling scaler.
+type FastBoxSampling struct {
+}
+
+// Scale implements the draw.Scaler interface.
+func (f *FastBoxSampling) Scale(dst draw.Image, dr image.Rectangle, src image.Image, sr image.Rectangle, op draw.Op, opts *draw.Options) {
+	ddx, ddy := dr.Dx(), dr.Dy()
+	sdx, sdy := sr.Dx(), sr.Dy()
+	if sdx < ddx && sdy < ddy {
```
at-wat: `if sr.Dx() < dr.Dx() && sr.Dy() < dr.Dy() {` would be less confusing. Overhead is enough small.

pkg/io/video/scaler_cgo.go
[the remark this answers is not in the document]
lherman-cs: Ah, that makes sense.

PR #102, opened 2020-03-04
Add hardware accelerated VP8/9 encoder

at-wat:
Using libva.
VP8/9 encoding by libva on post-Kabylake CPU (that has hardware acceleration) is faster than VP8 by libvpx.

### Reference issue
fix #98

PR #106, opened 2020-03-05
Support codec specific parameters

### LINKED ISSUE #105: Pass codec specific encoding parameters
Maybe as `map[string]interface{}`?
`prop.Codec.Quality` can be dropped instead.

at-wat:
I'd like to introduce it to vaapi codec (#102), which has a lot of low layer parameters and no API to get library default parameters like openh264 and vpx.

### Reference issue
Fixes #105

### DIFF
--- a/pkg/prop/prop.go
+++ b/pkg/prop/prop.go
@@ -119,11 +119,9 @@ type Codec struct {
 	// Target bitrate in bps.
 	BitRate int
 
-	// Quolity of the encoding [0-9].
-	// Larger value results higher quality and higher CPU usage.
-	// It depends on the selected codec.
-	Quality int
-
 	// Expected interval of the keyframes in frames.
 	KeyFrameInterval int
+
+	// Library specific parameter struct defined by each codec package
+	CodecParams interface{}
 }
--- a/examples/codecparam/README.md
+++ b/examples/codecparam/README.md
@@ -1,0 +1,29 @@
+## Instructions
+
+### Download example
+
+```
+go get github.com/pion/mediadevices/examples/codecparam
+```
+
+### Open example page
+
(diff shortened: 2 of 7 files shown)


lherman-cs:
This looks good to me. Do you mind to also add a new example for this?

at-wat:
Sure, I'll add an example.

at-wat:
Added an example and default parameter getter to make it possible to add fields without breaking existing code.

pkg/prop/prop.go
```diff
@@ -119,11 +119,9 @@ type Codec struct {
...
-	// Quolity of the encoding [0-9].
-	// Larger value results higher quality and higher CPU usage.
-	// It depends on the selected codec.
-	Quality int
-
 	// Expected interval of the keyframes in frames.
 	KeyFrameInterval int
+
+	// Library specific parameter struct defined by each codec package
+	CodecParams interface{}
```
at-wat: maybe better to make it 
```go
interface{
  Default() error
}
```
to force having default value loader.

at-wat: But, it can't handle parameter mutation of vpx's VP8/9.
libvpx uses one config struct both for VP8 and VP9. Current implementation is better for this case.
```go
cp, err := vpx.NewVP8Param()
if err != nil {
  panic(err)
}
p.CodecParams = cp
```

```go
var cp vpx.Param // type must be added per VP8/VP9 to have LoadDefault interface.
if err := cp.LoadDefault(); err != nil {
  panic(err)
}
p.CodecParams = cp
```

Let me keep the current interface.

lherman-cs: I think I like the way it is. If `CodecParam`, the caller and the codec can assume that it'll use default values.

PR #108, opened 2020-03-12
Fallback codecs

### LINKED ISSUE #107: Fallback codec
Select a fallback codec implementation if the first one was failed to initialize.
For example with the case of vaapi (higher priority) and vpx, try vaapi first, and if the environment doesn't have video acceleration hardware, use vpx.

It could be something like:
```go
codec.Register(webrtc.VP8, codec.VideoEncoderFallbacks(
  codec.VideoEncoderBuilder(vaapi.NewVP8Encoder),
  codec.VideoEncoderBuilder(vpx.NewVP8Encoder),
)
```

One problem is that user code doesn't have a way to know which implementation is used, so it's difficult to pass codec specific parameter if using fallback codec.
Adding `prop.Codec.ImplementationName string` and pass multiple `CodecParams` for each `ImplementationName` as `map[string]interface{}`?

at-wat:
Add a wrapper function to initialize codec and try next fallback codec if one fails.

### Reference issue
Fixes #107

### DIFF
--- a/examples/fallbackcodec/README.md
+++ b/examples/fallbackcodec/README.md
@@ -1,0 +1,29 @@
+## Instructions
+
+### Download example
+
+```
+go get github.com/pion/mediadevices/examples/fallbackcodec
+```
+
+### Open example page
+
+[jsfiddle.net](https://jsfiddle.net/z7ms3u5r/) you should see two text-areas and a 'Start Session' button
+
+### Run simple with your browsers SessionDescription as stdin
+
+In the jsfiddle the top textarea is your browser, copy that and:
+
+#### Linux
+
+Run `echo $BROWSER_SDP | simple`
+
+### Input simple's SessionDescription into your browser
+
+Copy the text that `simple` just emitted and copy into second text area
+
+### Hit 'Start Session' in jsfiddle, enjoy your video!
+
+A video should start playing in your browser above the input boxes, and will continue playing until you close the application.
+
+Congrats, you have used pion-WebRTC! Now start building something cool
--- a/examples/fallbackcodec/main.go
+++ b/examples/fallbackcodec/main.go
@@ -1,0 +1,149 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/pion/mediadevices"
+	"github.com/pion/mediadevices/examples/internal/signal"
+	"github.com/pion/mediadevices/pkg/codec"
+	"github.com/pion/mediadevices/pkg/frame"
+	"github.com/pion/webrtc/v2"
+
+	_ "github.com/pion/mediadevices/pkg/codec/opus" // This is required to register opus audio encoder
+	"github.com/pion/mediadevices/pkg/codec/vaapi"  // Load hardware accelerated codecs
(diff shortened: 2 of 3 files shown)


lherman-cs:
I feel like we need to redesign the codec package. This change definitely works. But, it feels that there's a disconnection between the library and the users.

1. We suddenly require them to register the codecs manually when they want to have a fallback

```
codec.Register(webrtc.VP8, codec.VideoEncoderFallbacks(
		// Use if hardware acceleration is available.
		codec.NamedVideoCodec{
			Name:  "vaapi",
			Codec: codec.VideoEncoderBuilder(vaapi.NewVP8Encoder),
		},
		// Software implementation should always available.
		codec.NamedVideoCodec{
			Name:  "vpx",
			Codec: codec.VideoEncoderBuilder(vpx.NewVP8Encoder),
		},
	))
```

2. Then, when they want to set the codec specific params, we require them to use the names from 1 (this also feels redundant).

```
		c.CodecParams = map[string]interface{}{
				"vaapi": vpxParam,
				"vpx":   vpxParam,
			}
```

3. Also, it's been really weird to me that when they import one of the codec packages, it suddenly registers with one or more payload types:

```
const (
	PCMU = "PCMU"
	PCMA = "PCMA"
	G722 = "G722"
	Opus = "OPUS"
	VP8  = "VP8"
	VP9  = "VP9"
	H264 = "H264"
)
```

I think we should instead let the users build the codecs by themselves, and we'll simply take them in `GetUserMedia`. What do you think?

at-wat:
> I think we should instead let the users build the codecs by themselves, and we'll simply take them in `GetUserMedia`.

I'm not very sure what you suggested.
Do you mean removing `codec.Register()` from `init()` of each codec package and register required codecs by the user code?

lherman-cs:
Sorry, it was definitely very vague. But, yes, you're right. I was thinking that we should remove `codec.Register()` from init and let the user to explicitly specify the codec in the `GetUserMedia`. 

I think it's really hard to get my idea across with just plain English ^^. So, I've created an experiment here, https://github.com/pion/mediadevices/commit/f6905fe81aba0f266b8170fadcd0cd8e954b86e4. It's definitely not a working example since I didn't implement everything, but I made changes that I think it's enough to express my idea. 

Following are the main changes:

1. We no longer require the user to specify the Codec Name, pass specific codec params, and import the codec to implicitly register the codec separately anymore. But, we rather have everything in `<specific codec>.Params`, e.g. `x264.Params`.
2. Specifying specific codec params becomes really easy and more straightforward since you just need to call `x264.NewParams()` and that's the only way you can specify the codec in `GetUserMedia`, so it's less error-prone. Also, we can avoid an empty interface (which is pretty awesome!)
3. Since now you're passing the builders, we can have it as an array and multiple builders with the same codec name will be fine. Essentially, we'll get a fallback method automatically.

at-wat:
OK. https://github.com/pion/mediadevices/commit/f6905fe81aba0f266b8170fadcd0cd8e954b86e4 definitely makes sense to me.

PR #112, opened 2020-03-17
Redesign codec

lherman-cs:
#### Description
1. We no longer require the user to specify the Codec Name, pass specific codec params, and import the codec to implicitly register the codec separately anymore. But, we rather have everything in `<specific codec>.Params`, e.g. `x264.Params`.
2. Specifying specific codec params becomes really easy and more straightforward since you just need to call `x264.NewParams()` and that's the only way you can specify the codec in `GetUserMedia`, so it's less error-prone. Also, we can avoid an empty interface (which is pretty awesome!)
3. Since now you're passing the builders, we can have it as an array and multiple builders with the same codec name will be fine. Essentially, we'll get a fallback method automatically.

#### Reference issue
#108, #112


examples/codecparam/main.go
```diff
@@ -5,18 +5,16 @@ import (
...
 
-	_ "github.com/pion/mediadevices/pkg/codec/opus" // This is required to register opus audio encoder
-	"github.com/pion/mediadevices/pkg/codec/vaapi"  // This is required to register VP8/VP9 video encoder
+	"github.com/pion/mediadevices/pkg/codec/vaapi" // This is required to register VP8/VP9 video encoder
 
-	// Note: If you don't have a camera or microphone or your adapters are not supported,
+	// Note: If you don't have a camera or your adapters are not supported,
 	//       you can always swap your adapters with our dummy adapters below.
 	// _ "github.com/pion/mediadevices/pkg/driver/videotest"
-	// _ "github.com/pion/mediadevices/pkg/driver/audiotest"
```
at-wat: The JS code on jsfiddle expects that the peer has both audio and video.
It might be better to have video only version of demo page.

at-wat: Opened an issue: https://github.com/pion/mediadevices/issues/115

pkg/codec/codec.go
```diff
@@ -8,5 +8,35 @@ import (
...
+	// Name represents the codec name
+	Name() string
+	// BuildVideoEncoder builds video encoder by given media params and video input
+	BuildVideoEncoder(r video.Reader, p prop.Media) (io.ReadCloser, error)
+}
+
+// BaseParams represents an codec's encoding properties
+type BaseParams struct {
+	// Target bitrate in bps.
+	BitRate int
```
at-wat: One concern is that it is not enforced to embed `codec.BaseParams` to `EncoderBuilder`.
It would be nice to have a unified method to set bit-rate and force key-framing. (It will be required to dynamically set bit-rate to realize WebRTC's keyframe request and bandwidth control.)

For example,
```go
type StreamReader interface {
	io.ReadCloser
	SetBitRate(int) error
	ForceKeyFrame() error
}

BuildVideoEncoder(r video.Reader, p prop.Media) (StreamReader, error)
```

It may be better to be done in different PR.

examples/codecparam/main.go
```diff
@@ -5,18 +5,16 @@ import (
 
 	"github.com/pion/mediadevices"
 	"github.com/pion/mediadevices/examples/internal/signal"
+	"github.com/pion/mediadevices/pkg/codec"
 	"github.com/pion/mediadevices/pkg/frame"
 	"github.com/pion/webrtc/v2"
 
-	_ "github.com/pion/mediadevices/pkg/codec/opus" // This is required to register opus audio encoder
-	"github.com/pion/mediadevices/pkg/codec/vaapi"  // This is required to register VP8/VP9 video encoder
+	"github.com/pion/mediadevices/pkg/codec/vaapi" // This is required to register VP8/VP9 video encoder
```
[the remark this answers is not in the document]
lherman-cs: ah, that's true

examples/codecparam/main.go
lherman-cs: Yep, that makes sense

pkg/codec/codec.go
lherman-cs: Yep, this is definitely the vision that I had too. And, yes, I think we should separate it into a different PR.

examples/codecparam/main.go
at-wat: https://jsfiddle.net/gh/get/library/pure/pion/mediadevices/tree/master/examples/internal/jsfiddle/video is available now.

lherman-cs:
@at-wat I've updated the PR.

PR #117, opened 2020-03-21
codec/vaapi: calculate max bit-rate from BaseParam

at-wat:
~branched from #112~ rebased


at-wat:
Rebased to master.

@lherman-cs please take a look.

PR #121, opened 2020-03-26
Add facedetection example

lherman-cs:
I think our repo is lacking in application examples or use cases. So, I figured that I should create an example to show how easy it is to do a simple machine learning application to detect faces. 

Another thing that I noticed from adding this example is that we need to organize our examples better and reduce the number of examples somehow so that it's more friendly to newcomers.


lherman-cs:
> It would be nice to add empty `go.mod` under `examples` to exclude example deps from top-level `go.mod`.
> https://github.com/golang/go/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file

Wow, I didn't know that I could do that. TIL.

lherman-cs:
@at-wat I've updated the `go.mod`

PR #123, opened 2020-03-26
Add bitrate measurement

lherman-cs:
This helper is going to be useful for writing unit tests for codec package, especially with rate limiting feature.


internal/codec/measurement_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestMeasureBitRateStatic(t *testing.T) {
+	r, w := io.Pipe()
+	dur := time.Second * 5
+	dataSize := 1000
+	var precision float64 = 8 // 1 byte
+
+	go func() {
+		data := make([]byte, dataSize)
+		for i := time.Second; i < dur; i += time.Second {
```
at-wat: Timing of actual goroutine execution is not guaranteed to be very immediate.
I think it's more stable to put `MeasureBitRate` into routine and iterate `Write` in main routine until `MeasureBitRate` actually ending.

internal/codec/measurement.go
```diff
@@ -0,0 +1,43 @@
...
+		n, err = r.Read(buf)
+		if err != nil {
+			if e, ok := err.(*mio.InsufficientBufferError); ok {
+				buf = make([]byte, 2*e.RequiredSize)
+				continue
+			}
+
+			if err == io.EOF {
+				totalBytes += n
+				fmt.Println("EOFF")
```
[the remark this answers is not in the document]
lherman-cs: oops

internal/codec/measurement_test.go
lherman-cs: Ah, that's a good point. I'll move it to the main routine, and probably add some synchronizations to make it more accurate.

internal/codec/measurement_test.go
```diff
@@ -0,0 +1,97 @@
...
+		wg.Done()
+		ticker := time.NewTicker(time.Second)
+
+		for {
+			select {
+			case <-ticker.C:
+				w.Write(data)
+			case <-done:
+				w.Close()
+				break
```
at-wat: It breaks inner select block and continue outer for loop.
```suggestion
				return
```

lherman-cs: ouch, my bad...

PR #125, opened 2020-03-27
codec/vaapi: return error on unsupported environment

at-wat:
Make it buildable on unsupported environment.
Return error on runtime.


at-wat:
It doesn't work as expected.
CGO enabled environment builds .c files (libva.h is required), and CGO disabled environment doesn't allow to have .c file.

PR #126, opened 2020-03-27
codec/vaapi: return error on unsupported environment

at-wat:
Make it buildable on unsupported environment.
Return error on runtime.


pkg/codec/vaapi/helper.c
```diff
@@ -1,3 +1,5 @@
+#ifndef NO_VAAPI
```
at-wat:
```suggestion
#ifdef HAS_VAAPI
```

PR #127, opened 2020-03-28
Tweak examples/go.mod

at-wat:


at-wat:
Without `replace`, build under examples directory uses latest master version of the mediadevices package even if the local tree has updates.
I'll update the commit message.

at-wat:
updated the commit message.

PR #129, opened 2020-03-29
driver/screen: support X11 16 bit color frame buffer

### LINKED ISSUE #120: Support more pixel format in x11 screen capture driver
Current code assumes that the pixel format is 32bit RGBA.

at-wat:
### Reference issue
Fixes #120

### DIFF
--- a/pkg/driver/screen/x11capture_linux.go
+++ b/pkg/driver/screen/x11capture_linux.go
@@ -7,7 +7,8 @@ package screen
 // #define XUTIL_DEFINE_FUNCTIONS
 // #include <X11/Xutil.h>
 // #include <X11/extensions/XShm.h>
-// void copyRGBA(void *dst, char *src, size_t l) { // 64bit aligned copy
+//
+// void copyBGR24(void *dst, char *src, size_t l) { // 64bit aligned copy
 //   uint64_t *d = (uint64_t*)dst;
 //   uint64_t *s = (uint64_t*)src;
 //   l /= 8;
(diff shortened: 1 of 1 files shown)


pkg/driver/screen/x11capture_linux.go
```diff
@@ -136,6 +198,17 @@ func newShmImage(dp *C.Display, screen int) (*shmImage, error) {
...
 	s := &shmImage{dp: dp}
 
+	switch {
+	case v.red_mask == 0xFF0000 && v.green_mask == 0xFF00 && v.blue_mask == 0xFF:
+		s.pixFmt = pixFmtBGR24
+	case v.red_mask == 0xF800 && v.green_mask == 0x7E0 && v.blue_mask == 0x1F:
+		s.pixFmt = pixFmtBGR16
+	default:
+		fmt.Printf("x11capture: unsupported pixel format (R: %0x, G: %0x, B: %0x)\n",
+			v.red_mask, v.green_mask, v.blue_mask)
```
at-wat: Error returned from here is silently discarded by `mediaDevices.selectVideo()`. To notify why no driver was selected, output information to stdout.

at-wat:
It can be tested using virtual X11 screen.

```shell
$ Xvfb :10 -ac -screen 0 800x600x16
$ xhost +
```
```shell
$ DISPLAY=:10 go run .
```

at-wat:
`pixFmtBGR24` is completely same as what I named RGBA.
Actual display frame buffer never has alpha value and just a padded to align 64 bits.

lherman-cs:
@at-wat Ah I see, my bad ^^. This looks good to me then!

PR #134, opened 2020-03-31
Fix wrong unit bitrate

lherman-cs:


pkg/codec/x264/x264.go
```diff
@@ -67,6 +67,10 @@ func newEncoder(r video.Reader, p prop.Media, params Params) (codec.ReadCloser,
 		params.KeyFrameInterval = 60
 	}
 
+	// Convert from bit/s to kbit/s because x264 uses kbit/s instead.
+	// Reference: https://code.videolan.org/videolan/x264/-/blob/7923c5818b50a3d8816eed222a7c43b418a73b36/encoder/encoder.c#L657
```
at-wat, line 71: This link seems not related to the bit-rate?

lherman-cs: Oops, it's a wrong file

PR #140, opened 2020-04-18
Refractor track to be more DRY

lherman-cs:
Resolves https://github.com/pion/mediadevices/issues/141

Before:
Screenshot from 2020-04-18 21-22-31

After:
Screenshot from 2020-04-18 21-22-25

### DIFF
--- a/sampler.go
+++ b/sampler.go
@@ -1,31 +1,35 @@
 package mediadevices
 
 import (
+	"math"
 	"time"
 
 	"github.com/pion/webrtc/v2/pkg/media"
 )
 
-type sampler struct {
-	track         LocalTrack
-	clockRate     float64
-	lastTimestamp time.Time
+type samplerFunc func(b []byte) error
+
+// newVideoSampler creates a video sampler that uses the actual video frame rate and
+// the codec's clock rate to come up with a duration for each sample.
+func newVideoSampler(t LocalTrack) samplerFunc {
+	clockRate := float64(t.Codec().ClockRate)
+	lastTimestamp := time.Now()
+
+	return samplerFunc(func(b []byte) error {
+		now := time.Now()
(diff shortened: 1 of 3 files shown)


sampler.go
```diff
@@ -6,26 +6,29 @@ import (
...
 
-func (s *sampler) sample(b []byte) error {
-	now := time.Now()
-	duration := now.Sub(s.lastTimestamp).Seconds()
-	samples := uint32(s.clockRate * duration)
-	s.lastTimestamp = now
+	return samplerFunc(func(b []byte) error {
+		now := time.Now()
+		duration := now.Sub(lastTimestamp).Seconds()
+		samples := uint32(clockRate * duration)
```
at-wat, line 20: I think this should be rounded.

sampler.go
```diff
@@ -6,26 +6,29 @@ import (
...
+		return t.WriteSample(media.Sample{Data: b, Samples: samples})
+	})
+}
 
-	sample := media.Sample{Data: b, Samples: samples}
-	return s.track.WriteSample(sample)
+// newAudioSampler creates a audio sampler that uses a fixed latency and
+// the codec's clock rate to come up with a duration for each sample.
+func newAudioSampler(t LocalTrack, latency time.Duration) samplerFunc {
+	samples := uint32(float64(t.Codec().ClockRate) * latency.Seconds())
```
at-wat, line 30: ditto

sampler.go
lherman-cs: makes sense. I'll update it.

sampler.go
lherman-cs: same

lherman-cs:
@at-wat Updated the PR. PTAL.

Issue #145, opened 2020-04-19
Redesign GetUserMedia API


lherman-cs:
As of now, `GetUserMedia` accepts a single parameter, `MediaStreamConstraints` defined as follows:

```golang
func (m *mediaDevices) GetUserMedia(constraints MediaStreamConstraints) (MediaStream, error) {
   ...
}

type MediaStreamConstraints struct {
	Audio MediaOption
	Video MediaOption
}

// MediaTrackConstraints represents https://w3c.github.io/mediacapture-main/#dom-mediatrackconstraints
type MediaTrackConstraints struct {
	prop.Media
	Enabled bool
	// VideoEncoderBuilders are codec builders that are used for encoding the video
	// and later being used for sending the appropriate RTP payload type.
	//
	// If one encoder builder fails to build the codec, the next builder will be used,
	// repeating until a codec builds. If no builders build successfully, an error is returned.
	VideoEncoderBuilders []codec.VideoEncoderBuilder
	// AudioEncoderBuilders are codec builders that are used for encoding the audio
	// and later being used for sending the appropriate RTP payload type.
	//
	// If one encoder builder fails to build the codec, the next builder will be used,
	// repeating until a codec builds. If no builders build successfully, an error is returned.
	AudioEncoderBuilders []codec.AudioEncoderBuilder
	// VideoTransform will be used to transform the video that's coming from the driver.
	// So, basically it'll look like following: driver -> VideoTransform -> codec
	VideoTransform video.TransformFunc
	// AudioTransform will be used to transform the audio that's coming from the driver.
	// So, basically it'll look like following: driver -> AudioTransform -> code
	AudioTransform audio.TransformFunc
}

type MediaOption func(*MediaTrackConstraints)
```

From the type definitions above, we see that we're using `MediaTrackConstraints` for unrelated stuff such as:
* VideoEncoderBuilders
* AudioEncoderBuilders
* VideoTransform
* AudioTransform

I think we should somehow move them away from `MediaTrackConstraints` because,
1. It's less confusing for the API user
2. Also, it'll make easier later when we want to interop with JS

The purpose of this issue thread is to talk about possible designs that can solve the problems above.

lherman-cs:
Implemented in https://github.com/pion/mediadevices/commit/2f5e4ee9148194d1be6519e88444ba518011d5b6

PR #149, opened 2020-04-20
Redesign GetUserMedia API

lherman-cs:
Resolves https://github.com/pion/mediadevices/issues/145

* Move encoder builders and transformers to MediaDevicesOption
* Remove Enable from constraints
* Implicitly disable a media when nil


examples/facedetection/main.go
```diff
@@ -56,22 +55,22 @@ func main() {
...
 	vp8Params, err := vpx.NewVP8Params()
 	if err != nil {
 		panic(err)
 	}
 	vp8Params.BitRate = 100000 // 100kbps
 
+	md := mediadevices.NewMediaDevices(
+		peerConnection,
+		mediadevices.WithVideoEncoders(&vp8Params),
+		mediadevices.WithVideoTransformers(markFacesTransformer),
```
at-wat, line 67: I think Transforms are not `MediaDevices` wide setting, but are media specific one.
For example, user may have multiple cameras and may want to apply transform to one of them.

lherman-cs: yea... that's true. Oh well, let me redesign it again.

PR #168, opened 2020-05-21
Support ranged/exact/oneof media constraints

### LINKED ISSUE #151: Support ranged media constraints
Pixel format is scored by complete match at now.
For example, I want to select one of `YUY2` and `UYVY` but not want to use `JPEG`.

https://github.com/pion/mediadevices/blob/949e850239b347a0750f005975caec7b8b77b5bb/pkg/prop/prop.go#L99-L112

at-wat:
### Reference issue
closes #151

### DIFF
--- a/pkg/prop/int.go
+++ b/pkg/prop/int.go
@@ -4,21 +4,28 @@ import (
 	"math"
 )
 
+// IntConstraint is an interface to represent integer value constraint.
 type IntConstraint interface {
 	Compare(int) (float64, bool)
 	Value() (int, bool)
 }
 
+// Int specifies ideal int value.
+// Any value may be selected, but closest value takes priority.
 type Int int
 
+// Compare implements IntConstraint.
 func (i Int) Compare(a int) (float64, bool) {
 	return math.Abs(float64(a-int(i))) / math.Max(math.Abs(float64(a)), math.Abs(float64(i))), true
 }
 
+// Value implements IntConstraint.
 func (i Int) Value() (int, bool) { return int(i), true }
 
+// IntExact specifies exact int value.
 type IntExact int
 
+// Compare implements IntConstraint.
 func (i IntExact) Compare(a int) (float64, bool) {
 	if int(i) == a {
 		return 0.0, true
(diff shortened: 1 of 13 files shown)


pkg/prop/int.go
```diff
@@ -0,0 +1,73 @@
...
+			return 0.0, true
+		}
+	}
+	return 1.0, false
+}
+
+type IntRanged struct {
+	Min   *int
+	Max   *int
+	Ideal *int
```
at-wat, line 40: We are assuming zero value of constraint as not-set.

lherman-cs:
@at-wat This is awesome work! It's kinda sad that there are many duplicated codes. But, I don't have an alternative either at this moment 😃.

I'm not really sure how to represent `Value`. I'm concerned with the safeness. Let's say that we want to get a value from an `IntConstraints`, the only safest way to get the value is to check first if the type is either `Int` or`IntExact`. But, this seems to be too implicit.

Maybe we can add a bool at the return value to tell the caller if the value can be represented? Similar to how go map getter works, `value, ok := someMap[key]`. If we take, `IntOneOf` as an example, it would be something like below:

```go
func (IntOneOf) Value() (int, bool) {
  return 0, false
}
```

What do you think?

at-wat:
`Value() (int, bool)` looks better. Thanks!

pkg/codec/vaapi/vp8.go
```diff
@@ -107,17 +107,32 @@ type encoderVP8 struct {
...
+	)
+	if width, ok = p.Width.Value(); !ok {
+		return nil, errors.New("media width is not constrained")
+	}
+	if height, ok = p.Height.Value(); !ok {
+		return nil, errors.New("media height is not constrained")
+	}
+	if frameRate, ok = p.FrameRate.Value(); !ok {
+		return nil, errors.New("media frame rate is not constrained")
+	}
```
at-wat, line 123: Maybe better to separate `prop.Media` (exact values) and `prop.MediaConstraint` (constraints)

PR #170, opened 2020-05-24
Select by DeviceID using StringConstraint

at-wat:
Caused panic at https://github.com/pion/mediadevices/blob/27d966611e1766d574ae78940df7c25656aaf428/pkg/prop/prop.go#L69-L70 when DeviceID was specified.

Compare DeviceID field on `FitnessDistance()` which was directly selected at upper layer.

### Reference issue
follows up #168


at-wat:
DeviceID field maybe better to support OneOf constraint.

at-wat:
working as expected now

PR #172, opened 2020-05-31
Harden driver's property requirement check

lherman-cs:
Make sure that the property that's given to the adapter is matched to one of the properties that the adapter provided. If none matched, the state would change to Close.


at-wat:
This depends on #173.

lherman-cs:
It's still not clear how to achieve this with the new `prop.MediaConstraints`. I'll close this for now and revisit again in the future.

PR #173, opened 2020-05-31
Fix invalid constraints merging

lherman-cs:
the constraints return by `selectBestDriver` should not be set to what the user specified, but instead the best property returned by the driver.

### DIFF
--- a/mediadevices.go
+++ b/mediadevices.go
@@ -217,8 +217,9 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
 		return nil, MediaTrackConstraints{}, errNotFound
 	}
 
-	constraints.selectedMedia = bestProp
-	constraints.selectedMedia.Merge(constraints.MediaConstraints)
+	constraints.selectedMedia = prop.Media{}
+	constraints.selectedMedia.MergeConstraints(constraints.MediaConstraints)
+	constraints.selectedMedia.Merge(bestProp)
 	return bestDriver, constraints, nil
 }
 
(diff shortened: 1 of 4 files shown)


mediadevices.go
```diff
@@ -218,7 +218,6 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
 	}
 
 	constraints.selectedMedia = bestProp
-	constraints.selectedMedia.Merge(constraints.MediaConstraints)
```
at-wat, line 221: I think it is needed for the parameters which can be given from continuous range of the value.
Currently the main usage is FrameRate of screen capture driver.

lherman-cs: Yep, that makes sense.

lherman-cs:
@at-wat just updated. PTAL. It's still not perfect and it feels off, but I think it's good for now to at least fix the issue. 

We should definitely consider to refractor the code base to use `prop.MediaConstraints` vs `prop.Media` properly. I've created https://github.com/pion/mediadevices/issues/178 to track on this.

mediadevices.go
```diff
@@ -217,8 +217,8 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
 		return nil, MediaTrackConstraints{}, errNotFound
 	}
 
-	constraints.selectedMedia = bestProp
-	constraints.selectedMedia.Merge(constraints.MediaConstraints)
+	constraints.selectedMedia.MergeConstraints(constraints.MediaConstraints)
```
at-wat, line 220: It would be safer to clear `constraints.selectedMedia` by `prop.Media{}` before merge.

lherman-cs: Yep, that makes sense.

PR #174, opened 2020-05-31
Add darwin camera support

lherman-cs:
* Add avfoundation Go and C bindings
* Add darwin camera adapter

### DIFF
--- a/pkg/avfoundation/AVFoundationBind/AVFoundationBind.h
+++ b/pkg/avfoundation/AVFoundationBind/AVFoundationBind.h
@@ -1,0 +1,78 @@
+// MIT License
+// 
+// Copyright (c) 2019-2020 Pion
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#pragma once
+
+#include <stddef.h>
+
+#define MAX_DEVICES                      8
+#define MAX_PROPERTIES                   64
+#define MAX_DEVICE_UID_CHARS             64
+
+typedef const char* STATUS;
+static STATUS STATUS_OK                       = (STATUS) NULL;
+static STATUS STATUS_NULL_ARG                 = (STATUS) "One of the arguments was null";
+static STATUS STATUS_DEVICE_INIT_FAILED       = (STATUS) "Failed to init device";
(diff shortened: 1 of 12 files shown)


at-wat:
Objective-C code seems be directly placed under the go module directory with filename `*.c` and
```c
#cgo CFLAGS: -x objective-c
#cgo LDFLAGS: -framework AVFoundation
```
ref: https://coderwall.com/p/l9jr5a/accessing-cocoa-objective-c-from-go-with-cgo

pkg/driver/camera/camera_darwin.go
```diff
@@ -0,0 +1,71 @@
+package camera
```
at-wat, line 1: all these MacOS specific code requires `// +build darwin`

lherman-cs:
> Objective-C code seems be directly placed under the go module directory with filename `*.c` and
> 
> ```c
> #cgo CFLAGS: -x objective-c
> #cgo LDFLAGS: -framework AVFoundation
> ```
> 
> ref: https://coderwall.com/p/l9jr5a/accessing-cocoa-objective-c-from-go-with-cgo

@at-wat I really thought that I tried that and it didn't work... 😅. But, anyhow, I just updated the PR with that approach. Although, I didn't put the Objective C source code in the same folder, I'd like to preserve the metadata from XCode so that other developers can easily jump in.

pkg/driver/camera/camera_darwin.go
lherman-cs: I think that should be automatic, https://golang.org/pkg/go/build/#hdr-Build_Constraints. Since I named the file with `_darwin` prefix, it should be only for darwin?

at-wat: Oops, yes, this file is fine.
But, `pkg/avfoundation/avfoundation_callback.go` needs that to avoid failing on `go build ./...`.

lherman-cs: Oops...

lherman-cs: @at-wat updated. PTAL

pkg/avfoundation/AVFoundationBind/AVFoundationBind.h
```diff
@@ -0,0 +1,64 @@
+//
+//  AVFoundationBind.h
+//  AVFoundationBind
+//
+//  Created by Herman, Lukas on 5/23/20.
+//  Copyright © 2020 Herman, Lukas. All rights reserved.
```
at-wat, line 6: Is it non-MIT license?

pkg/avfoundation/avfoundation_callback_darwin.go
```diff
@@ -0,0 +1,53 @@
...
+var handles = make(map[handleID]dataCb)
+
+type handleID int
+
+//export onData
+func onData(userData unsafe.Pointer, buf unsafe.Pointer, length C.int) {
+	data := C.GoBytes(buf, length)
+
+	handleNum := (*C.int)(userData)
+	cb := lookup(handleID(*handleNum))
```
at-wat, line 24: Looks better to handle `cb == nil` case. I guess it may happen right after `unregister`.

pkg/avfoundation/AVFoundationBind/AVFoundationBind.h
lherman-cs: Oops, sorry, it was just automatically generated by XCode. I'll fix that.

pkg/avfoundation/avfoundation_callback_darwin.go
lherman-cs: Yea, but, I'm not sure the right behavior for this. Since this lookup is to give us back the session instance, if this failed to lookup, we can't return the error back to the session object.

lherman-cs:
`I don't really know about Licensing. But, I've removed the License comments from the source codes. So, it should use the root's [LICENSE](https://github.com/pion/mediadevices/blob/master/LICENSE)?

at-wat:
> But, I've removed the License comments from the source codes. So, it should use the root's LICENSE?

In my understanding, MIT license doesn't have explicit policy about it.
Some of the other licenses like GPL explicitly requires to include short preamble of the license to each file. Apache license recommends (stated by `should`) to include such preamble, but FAQ said that [`Only one full copy of the license is needed per distribution.`](https://infra.apache.org/apply-license.html)

I think it's preferable to include a short preamble.

pkg/avfoundation/avfoundation_callback_darwin.go
at-wat: I think it may happens if the user unregister `handleID` and the data captured right before unregistering arrives here.
In this case `if cb == nil { return }` would be enough.

lherman-cs: Ah, I see what you mean. I'll add that. Thanks!

lherman-cs:
@at-wat updated the lookup check and licensing stuff.

PR #176, opened 2020-06-02
Renew LICENSE year

lherman-cs:


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Pion
+Copyright (c) 2020 Pion
```
at-wat, line 3: I think it's better to include first year to clarify how far back the license is applied.
```suggestion
Copyright (c) 2019-2020 Pion
```

lherman-cs: Ouch sorry, I'm pretty bad with licensing.

PR #179, opened 2020-06-06
Add raw audio decoder

lherman-cs:
* Add Int16Interleaved and Int16NonInterleaved formats
* Add Float32Interleaved and Float32NonInterleaved formats
* Add unit tests

### DIFF
--- a/pkg/wave/decoder_test.go
+++ b/pkg/wave/decoder_test.go
@@ -1,0 +1,314 @@
+package wave
+
+import (
+	"encoding/binary"
+	"math"
+	"reflect"
+	"testing"
+)
+
+func TestCalculateChunkInfo(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/wave/decoder_test.go
```diff
@@ -0,0 +1,313 @@
...
+			expected: ChunkInfo{
+				Len:          2,
+				Channels:     1,
+				SamplingRate: 0,
+			},
+			expectErr: false,
+		},
+	}
+
+	for testCaseName, testCase := range testCases {
```
at-wat, line 81:
```suggestion
	for testCaseName, testCase := range testCases {
		testCase := testCase
```
for parallel test safety.

lherman-cs: Ah, thanks. I forgot about that.

PR #180, opened 2020-06-07
Directly copy memory in audio decoder if endianness matches

at-wat:
https://github.com/pion/mediadevices/pull/179#pullrequestreview-425790764

### DIFF
--- a/pkg/wave/decoder.go
+++ b/pkg/wave/decoder.go
@@ -4,6 +4,8 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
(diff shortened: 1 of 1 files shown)


lherman-cs:
@at-wat this looks good! Btw, I created a PR to this branch, https://github.com/pion/mediadevices/pull/181. It's about adding benchmark for comparing host vs non-host endian, and a minor fix about non-interleaved formats.

at-wat:
rebased

at-wat:
#181 contains this with a fix

PR #182, opened 2020-06-08
Opus support int16

at-wat:
Usually int16 is sound system native format.

### DIFF
--- a/pkg/codec/opus/opus.go
+++ b/pkg/codec/opus/opus.go
@@ -1,6 +1,7 @@
 package opus
 
 import (
+	"errors"
 	"fmt"
 	"math"
 
@@ -9,11 +10,12 @@ import (
 	"github.com/pion/mediadevices/pkg/io/audio"
 	"github.com/pion/mediadevices/pkg/prop"
 	"github.com/pion/mediadevices/pkg/wave"
+	"github.com/pion/mediadevices/pkg/wave/mixer"
 )
 
 type encoder struct {
 	engine *opus.Encoder
-	inBuff *wave.Float32Interleaved
+	inBuff wave.Audio
 	reader audio.Reader
 }
 
@@ -32,6 +34,10 @@ func newEncoder(r audio.Reader, p prop.Media, params Params) (codec.ReadCloser, 
 		params.BitRate = 32000
 	}
 
+	if params.ChannelMixer == nil {
+		params.ChannelMixer = &mixer.MonoMixer{}
+	}
+
 	// Select the nearest supported latency
 	var targetLatency float64
 	// TODO: use p.Latency.Milliseconds() after Go 1.12 EOL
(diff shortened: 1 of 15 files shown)


pkg/codec/opus/mixer.go
```diff
@@ -0,0 +1,44 @@
...
+type MonoMixer struct {
+}
+
+func (m *MonoMixer) Mix(dst wave.Audio, src wave.Audio) error {
+	if dst.ChunkInfo().Len != src.ChunkInfo().Len {
+		return errors.New("buffer size mismatch")
+	}
+	dstSetter, ok := dst.(interface {
+		Set(i, ch int, s wave.Sample)
+	})
```
lherman-cs, line 24: I think we should define the interface similarly to how `image/draw` package does it. So, maybe something like:

```go
type Audio interface {
    wave.Audio
    Set(i, ch int, s wave.Sample)
}
```

pkg/codec/opus/opus.go
```diff
@@ -58,47 +61,98 @@ func newEncoder(r audio.Reader, p prop.Media, params Params) (codec.ReadCloser,
...
-			panic("unsupported audio format")
-		}
-		switch {
-		case b.Size.Channels == 1 && e.inBuff.ChunkInfo().Channels != 1:
-			for _, d := range b.Data {
-				for ch := 0; ch < e.inBuff.ChunkInfo().Channels; ch++ {
-					e.inBuff.Data = append(e.inBuff.Data, d)
-				}
+		switch b := buff.(type) {
+		case *wave.Float32Interleaved:
```
lherman-cs, line 99: I think we can move this logic to `io/audio` and make it similar to how we do it for video scaling. So, before we read from the reader, we transform it so that it will always return the right latency.

at-wat: Do you have a good English word to express this functionality that splits signal at uniform interval?

pkg/codec/opus/mixer.go
at-wat: Hmm, I don't have enough vocabulary for naming the package.
My current candidates are `wave/sound` (might be too general) and `wave/dsp` (sounds too high layer).

lherman-cs: Hmm, how about if we put it in `wave` still, but the interface is `EdittableAudio`?

pkg/codec/opus/opus.go
lherman-cs: hmm, not really sure either... How about `Buffer`? Since we would put in under `io/audio` and the signature should be clear that this is a transformation, I think it's clear for the caller. e.g.

```go
import "github.com/pion/mediadevices/pkg/io/audio"

// This will create a reader with a buffer of 20 millisecond for each read
r, err := audio.Buffer(time.Millisecond * 20)
```

pkg/codec/opus/mixer.go
at-wat: Yea, it sounds better.

PR #186, opened 2020-06-09
Make raw audio decoder more practical

lherman-cs:
Unfortunately, in audio world, there's no clear standard for defining formats like fourcc for video. While defining our own consts for audio is fine, it's impractical when we're trying to integrate frameworks like avfoundation.

### DIFF
--- a/pkg/wave/decoder.go
+++ b/pkg/wave/decoder.go
@@ -9,16 +9,30 @@ import (
 )
 
 // Format represents how audio is formatted in memory
-type Format string
+type Format fmt.Stringer
 
-const (
-	FormatInt16Interleaved      Format = "Int16Interleaved"
-	FormatInt16NonInterleaved          = "Int16NonInterleaved"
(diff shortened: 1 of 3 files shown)


pkg/wave/decoder.go
```diff
@@ -9,16 +9,28 @@ import (
 )
 
 // Format represents how audio is formatted in memory
-type Format string
+type Format struct {
```
at-wat, line 12: It seems to cover only constant bit-width raw audio.
Usually it's enough but I think it's better to keep it available for other kind of formats, for example compressed data, as same as FourCC. (`wave.Audio` can be implemented for compressed data.)

at-wat:
```go
type Format fmt.Stringer
```
this would be sufficient, I think.

lherman-cs: That's a good idea! Thanks. I've just updated the PR with that change.

lherman-cs:
> LGTM
> Sorry for the delay!

No worries. Thanks!

PR #206, opened 2020-08-20
Add V4L2_PIX_FMT_YUV420 support for Linux

lherman-cs:
According to https://www.kernel.org/doc/html/v4.10/media/uapi/v4l/pixfmt-yuv420.html, YU12 is equal to V4L2_PIX_FMT_YUV420 in v4l2 term. In fourcc, V4L2_PIX_FMT_YUV420 is equal to I420, V4L2_PIX_FMT_YUV420.


at-wat (approved):
LGTM

PR #209, opened 2020-09-03
Add frame Copier

lherman-cs:
This is an incremental effort to redesign the project. Frame copy is necessary for supporting a video broadcaster.

Benchmark:

```
goos: darwin
goarch: amd64
pkg: github.com/pion/mediadevices/pkg/io/video
BenchmarkCopyOptimized
BenchmarkCopyOptimized-4            9290            132430 ns/op             670 B/op          0 allocs/op
BenchmarkCopyNaive
BenchmarkCopyNaive-4                2613            446601 ns/op         3130664 B/op          4 allocs/op
PASS
ok      github.com/pion/mediadevices/pkg/io/video       5.769s
```

### DIFF
--- a/pkg/io/video/framebuffer.go
+++ b/pkg/io/video/framebuffer.go
@@ -1,0 +1,214 @@
+package video
+
+import (
+	"image"
+)
+
+// FrameBuffer is a buffer that can store any image format.
+type FrameBuffer struct {
+	buffer []uint8
+	tmp    image.Image
+}
+
+// NewFrameBuffer creates a new FrameBuffer instance and initialize internal buffer
+// with initialSize
+func NewFrameBuffer(initialSize int) *FrameBuffer {
+	return &FrameBuffer{
+		buffer: make([]uint8, initialSize),
+	}
+}
+
+func (buff *FrameBuffer) storeInOrder(srcs ...[]uint8) {
+	var neededSize int
+
+	for _, src := range srcs {
+		neededSize += len(src)
+	}
+
+	if len(buff.buffer) < neededSize {
+		if cap(buff.buffer) >= neededSize {
+			buff.buffer = buff.buffer[:neededSize]
+		} else {
+			buff.buffer = make([]uint8, neededSize)
+		}
+	}
+
+	var currentLen int
+	for _, src := range srcs {
+		copy(buff.buffer[currentLen:], src)
+		currentLen += len(src)
+	}
+}
+
+// Load loads the current owned image
+func (buff *FrameBuffer) Load() image.Image {
+	return buff.tmp
+}
+
+// StoreCopy makes a copy of src and store its copy. StoreCopy will reuse as much memory as it can
+// from the previous copies. For example, if StoreCopy is given an image that has the same resolution
+// and format from the previous call, StoreCopy will not allocate extra memory and only copy the content
+// from src to the previous buffer.
+func (buff *FrameBuffer) StoreCopy(src image.Image) {
+	switch src := src.(type) {
+	case *image.Alpha:
+		clone, ok := buff.tmp.(*image.Alpha)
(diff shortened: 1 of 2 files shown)


pkg/io/video/copy.go
```diff
@@ -0,0 +1,196 @@
...
+	var currentLen int
+	for _, src := range srcs {
+		copy(copier.buffer[currentLen:], src)
+		currentLen += len(src)
+	}
+}
+
+// Copy copies src to dst. Since dst can be nil, the caller is responsible to assign the
+// return value from Copy to dst. When dst is not nil, Copy will reuse its memory and only
+// copy values from src to dst.
```
at-wat, line 44: I think it's better to split shallow copy func and deep copy func with explicit names.

pkg/io/video/copy.go
```diff
@@ -0,0 +1,196 @@
...
+func (copier *Copier) Copy(dst, src image.Image) image.Image {
+	switch src := src.(type) {
+	case *image.Alpha:
+		clone, ok := dst.(*image.Alpha)
+		if ok {
+			*clone = *src
+		} else {
+			copied := *src
+			clone = &copied
```
at-wat, line 53: I feel it's better to return error if type of src and dst is not matched.

pkg/io/video/copy.go
lherman-cs: Not sure what you mean, but this function only does deep copy.

pkg/io/video/copy.go
lherman-cs: Why is that? The thing is if `dst` is nil, ok will be equal to false anyway. So, I think it'll only add extra codes and more overhead.

pkg/io/video/copy.go
at-wat: > When dst is not nil, Copy will reuse its memory and only copy values from src to dst.

This function looks doing deep copy only if dst is not nil.

pkg/io/video/copy.go
at-wat: I think if user try to copy `*image.CMYK` into `*image.RGBA`, it's misuse of the function.

pkg/io/video/copy.go
lherman-cs: Oh I see what you mean. So, I guess you're thinking about users misusing this by calling copy from the same Copier instance in different goroutines. I can definitely see that to be problematic.

lherman-cs: So, I think we'll create `ShallowCopy` and `DeepCopy`. But, should we allow `ShallowCopy` to have `nil` as its dst? Or should we just enforce users to not have `nil` at all as its dst?

pkg/io/video/copy.go
lherman-cs: Yep, makes sense. I think we'll just enforce this with panic since this is a misuse.

lherman-cs: After second thought, I'd like to keep the behaviour as it is. My motivation is that I want to use Copy for another function, "broadcast". So, it's important to keep it generic since users might change the image format at anytime in the pipeline.

lherman-cs:
@at-wat I think the naming is really confusing. I've updated by renaming it to `FrameBuffer` instead, and changed the API to be `Load` and `Store` style. This way, it'll let the users aware the buffer is not meant to be shared across multiple goroutines unless there's a synchronization in place.

The design is inspired by atomic.Value, https://golang.org/pkg/sync/atomic/#Value.Load.

What do you think about this approach?

at-wat:
`FrameBuffer` API sounds good to me!

lherman-cs:
@at-wat updated the PR, PTAL.

at-wat (approved):
LGTM

PR #212, opened 2020-09-07
Convert underlying wave buffer to be uint8

lherman-cs:
This follows image package from the standard library. By having a homogenous data type for storing the samples, it makes easier to manipulate the raw data in a generic way.

The main motivation of doing this is so that we can have ChunkBuffer similar to what we have for images, https://github.com/pion/mediadevices/pull/209. If we had different data types to store the samples, we need to have multiple slices with different data types or we need to use unsafe to remove the type safety (which is not recommended generally).

This approach is inspired by https://golang.org/src/image/image.go?s=12897:12943#L428.

#### Reference issue
Fixes #...


edaniels:
@lherman-cs do you still want to get this in?

PR #213, opened 2020-09-11
Add human readable driver not found error message

lherman-cs:
Many times users tried to run the example and found that it couldn't find a driver, the error message it has been confusing and too generic. This PR will let users know what the library can discover, this way we can isolate the issues a lot faster.

### DIFF
--- a/mediadevices.go
+++ b/mediadevices.go
@@ -3,6 +3,7 @@ package mediadevices
 import (
 	"fmt"
 	"math"
+	"strings"
 
 	"github.com/pion/mediadevices/pkg/driver"
 	"github.com/pion/mediadevices/pkg/prop"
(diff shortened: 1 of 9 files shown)


mediadevices.go
```diff
@@ -214,7 +215,24 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
...
-		return nil, MediaTrackConstraints{}, errNotFound
+		var foundProperties []string
+		for _, props := range driverProperties {
+			for _, p := range props {
+				foundProperties = append(foundProperties, fmt.Sprint(&p))
+			}
+		}
+
+		err := fmt.Errorf(`
+%s:
```
at-wat, line 226: `%w` would be better to allow error unwrapping.
https://golang.org/pkg/fmt/#Errorf

lherman-cs: Ah right, I'm still not used to it..

lherman-cs: @at-wat updated

mediadevices.go
```diff
@@ -214,7 +215,24 @@ func selectBestDriver(filter driver.FilterFn, constraints MediaTrackConstraints)
...
-		return nil, MediaTrackConstraints{}, errNotFound
+		var foundProperties []string
+		for _, props := range driverProperties {
+			for _, p := range props {
+				foundProperties = append(foundProperties, fmt.Sprint(&p))
+			}
+		}
+
+		err := fmt.Errorf(`
+%w:
```
at-wat, line 226: This will add empty line before the error description.
```suggestion
		err := fmt.Errorf(`%w:
```

at-wat (approved):
Added an additional minor comment. LGTM

mediadevices.go
lherman-cs: Ouch.., I'll fix it

PR #219, opened 2020-10-01
Add generic wave's Buffer

lherman-cs:
This is an incremental effort to redesign the project. Wave's buffer is necessary for supporting an audio broadcaster.

Similar to https://github.com/pion/mediadevices/pull/209, but for audio.


at-wat (approved):
LGTM

PR #221, opened 2020-10-02
Move current broadcast test to io

lherman-cs:
Changes:
* Move core broadcast tests to io
* Add targeted tests for video.Broadcast


at-wat (approved):
LGTM

PR #222, opened 2020-10-02
Add code coverage report

lherman-cs:


at-wat (commented):
Maybe better to use a codecov official action instead of the bash script.
https://github.com/codecov/codecov-action

lherman-cs:
> https://github.com/codecov/codecov-action

Good call!

at-wat (approved):
LGTM

PR #224, opened 2020-10-04
Fix unstored audio

lherman-cs:
* Add unit test to check internal buffer integrity
* Fix unstored audio in internal buffer

### DIFF
--- a/pkg/wave/buffer_test.go
+++ b/pkg/wave/buffer_test.go
@@ -1,8 +1,14 @@
 package wave
 
 import (
+	"errors"
+	"fmt"
 	"reflect"
 	"testing"
+)
+
+var (
+	errIdenticalAddress = errors.New("Cloned audio has the same memory address with the original audio")
 )
 
 func TestBufferStoreCopyAndLoad(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/wave/buffer_test.go
```diff
@@ -12,8 +12,9 @@ func TestBufferStoreCopyAndLoad(t *testing.T) {
 		SamplingRate: 48000,
 	}
 	testCases := map[string]struct {
-		New    func() EditableAudio
-		Update func(EditableAudio)
+		New      func() EditableAudio
+		Update   func(EditableAudio)
+		Validate func(Audio, Audio) bool
```
at-wat, line 17:
```suggestion
		Validate func(*testing.T, Audio, Audio) bool
```
would be better to make it easy to find why the validation is failed.

lherman-cs: I see. In that case, I'll just return error then instead, and let the higher level to print that.

at-wat: `t.Error` outputs the filename and line number where it errored. Calling `t.Error` in Validate func makes it easy to find the failing line.

lherman-cs: Oh! I didn't know that. I'll do that then!

lherman-cs: Updated.

at-wat (approved):
LGTM

PR #214, opened 2020-09-13
Add pull-based Broadcaster

lherman-cs:
Changes:
* Add generic io.Reader
* Add generic broadcaster
* Add specialize video broadcaster
* Add ring buffer to avoid frames being dropped

PR #225, opened 2020-10-04
Add audio pull-based broadcast

lherman-cs:
Similar to https://github.com/pion/mediadevices/pull/214, but for audio.

This PR depends on https://github.com/pion/mediadevices/pull/224.

### DIFF
--- a/pkg/io/audio/broadcast.go
+++ b/pkg/io/audio/broadcast.go
@@ -1,0 +1,76 @@
+package audio
+
+import (
+	"errors"
+
+	"github.com/pion/mediadevices/pkg/io"
+	"github.com/pion/mediadevices/pkg/wave"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/io/audio/broadcast.go
```diff
@@ -0,0 +1,76 @@
+package audio
+
+import (
+	"fmt"
+
+	"github.com/pion/mediadevices/pkg/io"
+	"github.com/pion/mediadevices/pkg/wave"
+)
+
+var errEmptySource = fmt.Errorf("Source can't be nil")
```
at-wat, line 10:
```suggestion
var errEmptySource = errors.New("source can't be nil")
```

lherman-cs: Done

at-wat (approved):
LGTM

PR #226, opened 2020-10-06
Remove webrtc dependency in codec and its sub packages

lherman-cs:
Depending on webrtc in codec is unnecessary. This PR is to clean up the dependency graph and make it more centralized, having the webrtc dependency in the root package only.

### DIFF
--- a/pkg/codec/codec.go
+++ b/pkg/codec/codec.go
@@ -8,6 +8,17 @@ import (
 	"github.com/pion/mediadevices/pkg/prop"
 )
 
+// Name represents codec official name. It's possible to have more than 1 implementations
+// for the same codec name, e.g. openh264 vs x264.
+type Name string
+
+const (
+	NameOpus Name = "opus"
+	NameH264 Name = "H264"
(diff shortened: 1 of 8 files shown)


pkg/codec/codec.go
```diff
@@ -8,14 +8,25 @@ import (
...
 
+// Name represents codec official name. It's possible to have more than 1 implementations
+// for the same codec name, e.g. openh264 vs x264.
+type Name string
+
+const (
+	NameOpus = "opus"
+	NameH264 = "H264"
+	NameVP8  = "VP8"
+	NameVP9  = "VP9"
```
at-wat, line 19:
```suggestion
	NameOpus Name = "opus"
	NameH264 Name = "H264"
	NameVP8  Name = "VP8"
	NameVP9  Name = "VP9"
```
would be more explicit.

lherman-cs: Ouch, I forgot...

at-wat (approved):
LGTM

PR #227, opened 2020-10-07
Replace Name with RTPCodec in codec builder

lherman-cs:
Allowing users to implement RTPCodec will give users freedom to have a custom encoder with custom RTP payload. This is another incremental update toward the new design.

Use case:

```go
package main

import (
	"github.com/pion/mediadevices"
	"github.com/pion/mediadevices/pkg/codec/openh264"
	"github.com/pion/mediadevices/pkg/codec/x264"
	"github.com/pion/mediadevices/pkg/prop"

	// Note: If you don't have a camera or microphone or your adapters are not supported,
	_ "github.com/pion/mediadevices/pkg/driver/camera" // This is required to register camera adapter
)

func main() {
	var setting webrtc.SettingEngine

	x264Params, _ := x264.NewParams()
	openh264Params, _ := openh264.NewParams()
	rtpTracker, _ := mediadevices.NewRTPTracker(x264Params, openh264Params) 
	rtpTracker.PopulateSetting(&setting) // <--- We can easily register custom RTPCodec

	// User can still modify setting to their needs here

	api := webrtc.NewAPI(webrtc.WithSettingEngine(&setting))

	// Assume that we have configured the api and have proper config
	pc, _ := api.NewPeerConnection(peerConnectionConfig)

	mediaStream, _ := mediadevices.GetUserMedia(mediadevices.MediaStreamConstraints{
		Video: func(constraint *mediadevices.MediaTrackConstraints) {
			constraint.Width = prop.Int(600)
			constraint.Height = prop.Int(400)
		},
	})

	for _, mediaTrack := range mediaStream.GetTracks() {
		// rtpTracker.Track will create LocalRTPTrack, which later can be used for pulling video/audio frames.
		pc.AddTransceiverFromTrack(rtpTracker.Track(mediaTrack),
			webrtc.RtpTransceiverInit{
				Direction: webrtc.RTPTransceiverDirectionSendonly,
			},
		)
	}

	// peerconnection should negotiate with the other peer
	// peerconnection should set the negotiated codec with LocalRTPTrack.setParameters()

	// mediadevices then will set the codec and attach the encoder to the pipeline. PeerConnection now can start
	// pulling encoded frames from mediadevices.
}
```

Note: I had to bring back webrtc dependency to codec because I don't know any better place to have the mapping from codec name to RTPCodec.

### DIFF
--- a/mediadevices_test.go
+++ b/mediadevices_test.go
@@ -161,8 +161,10 @@ type mockParams struct {
 	name string
 }
 
-func (params *mockParams) Name() codec.Name {
-	return codec.Name(params.name)
+func (params *mockParams) RTPCodec() *codec.RTPCodec {
+	rtpCodec := codec.NewRTPH264Codec(90000)
+	rtpCodec.Name = params.name
+	return rtpCodec
 }
 
 func (params *mockParams) BuildVideoEncoder(r video.Reader, p prop.Media) (codec.ReadCloser, error) {
--- a/pkg/codec/codec.go
+++ b/pkg/codec/codec.go
@@ -6,18 +6,31 @@ import (
 	"github.com/pion/mediadevices/pkg/io/audio"
 	"github.com/pion/mediadevices/pkg/io/video"
 	"github.com/pion/mediadevices/pkg/prop"
+	"github.com/pion/webrtc/v2"
 )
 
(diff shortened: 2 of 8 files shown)


at-wat (approved):
LGTM

PR #228, opened 2020-10-10
Add audio change detection

lherman-cs:
This allows us to not need pass audio metadata manually. Consequently, users who want to transform audio can implement their transformation easily.


at-wat (approved):
LGTM

PR #229, opened 2020-10-10
Use MediaDeviceInfo instead of webrtc.RTPCodecType

lherman-cs:
Changes:
  * Add unit tests for mediastream
  * Remove webrtc.RTPCodecType dependency in mediastream
  * Add Kind to Tracker interface


at-wat (approved):
LGTM

PR #231, opened 2020-10-11
Replace sampler algorithm to be more generic

lherman-cs:
Since RTP only cares about timestamps and each sample should increment the timestamp following the negotiated Clockrate, we don't need to calculate fps and use latency, we can simply measure the duration per sample.

### DIFF
--- a/track.go
+++ b/track.go
@@ -48,7 +48,6 @@ type track struct {
 func newTrack(opts *MediaDevicesOptions, d driver.Driver, constraints MediaTrackConstraints) (*track, error) {
 	var encoderBuilders []encoderBuilder
 	var rtpCodecs []*webrtc.RTPCodec
-	var buildSampler func(t LocalTrack) samplerFunc
 	var kind MediaDeviceType
 	var err error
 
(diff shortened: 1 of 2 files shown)


track.go
```diff
@@ -61,14 +60,10 @@ func newTrack(opts *MediaDevicesOptions, d driver.Driver, constraints MediaTrack
 	case driver.VideoRecorder:
 		kind = VideoInput
 		rtpCodecs = opts.codecs[webrtc.RTPCodecTypeVideo]
-		buildSampler = newVideoSampler
 		encoderBuilders, err = newVideoEncoderBuilders(r, constraints)
 	case driver.AudioRecorder:
 		kind = AudioInput
 		rtpCodecs = opts.codecs[webrtc.RTPCodecTypeAudio]
-		buildSampler = func(t LocalTrack) samplerFunc {
-			return newAudioSampler(t, constraints.selectedMedia.Latency)
```
at-wat, line 70: In my experience, using raw system time, which contains jitter, to RTP timestamp causes problem on playback, and it required low pass filter to use system time based duration. (It might work without visible problem if CPU usage is enough low.)

lherman-cs: Oh that's good to know! My issue currently is that we don't have a good way to propagate the latency information. We currently got it from the selected property, but the latency itself can be dynamic and changed in transformations, which lead to an incorrect value.

lherman-cs: Or maybe we can use DetectChanges, https://github.com/pion/mediadevices/blob/master/pkg/io/audio/detect.go#L12, to calculate the latency using the chunk length and sample rate?

lherman-cs:
Closing since the approach can create noise/jitter.

PR #232, opened 2020-10-11
Switch from type alias to embedded struct

lherman-cs:
Embedded struct provides more future compatibility than type alias.


at-wat (approved):
LGTM

PR #235, opened 2020-10-14
Add meta reader helper

lherman-cs:
Since broadcaster has a ring buffer, we can take advantage of this property to read the meta data for video/audio without losing any data.

### DIFF
--- a/meta.go
+++ b/meta.go
@@ -1,0 +1,35 @@
+package mediadevices
+
+import (
+	"github.com/pion/mediadevices/pkg/io/audio"
+	"github.com/pion/mediadevices/pkg/io/video"
+	"github.com/pion/mediadevices/pkg/prop"
+)
+
+// detectCurrentVideoProp is a small helper to get current video property
+func detectCurrentVideoProp(broadcaster *video.Broadcaster) (prop.Media, error) {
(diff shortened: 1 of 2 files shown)


at-wat:
As it requires polling to detect prop update, some frames between source prop update and next polling timing will be handled with old prop.
I guess WebRTC implementation in browsers immediately reflect source prop change to encoding setting. (not very sure)

at-wat (commented):
As it requires polling to detect prop update, some frames between source prop update and next polling timing will be handled with old prop.
I guess WebRTC implementation in browsers immediately reflect source prop change to encoding setting. (not very sure)

lherman-cs:
> As it requires polling to detect prop update, some frames between source prop update and next polling timing will be handled with old prop.
> I guess WebRTC implementation in browsers immediately reflect source prop change to encoding setting. (not very sure)

I was planning to use these helpers to initialize the encoders only. In the future, I think our encoders should have a way to update the parameters on the fly. When we have this, we'll wrap the reader with video.DetectChanges, which will always inform us whenever any of the required parameters change, and we'll simply use the encoder's API to update its parameters.

at-wat (approved):
LGTM

PR #237, opened 2020-10-15
Remove buffer in linux camera

lherman-cs:
Speed up camera latency. On raspberry pi 3, the latency goes down from 3-5 seconds to <1 second with this change.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -83,6 +83,8 @@ func (c *camera) Open() error {
 		return err
 	}
 
+	// Late frames should be discarded. Buffering should be handled in higher level.
+	cam.SetBufferCount(1)
 	c.cam = cam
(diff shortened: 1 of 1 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -83,6 +83,8 @@ func (c *camera) Open() error {
 		return err
 	}
 
+	// Late frames should be discarded. Buffering should be handled in higher level.
+	cam.SetBufferCount(1)
```
at-wat, line 87: I guess dual buffer is required to load next frame during using previous frame. (not very sure it's required)

lherman-cs: From my test, this seems to be fine. I also tried with double buffer, but it has higher latency.

at-wat (approved):
LGTM

PR #240, opened 2020-10-26
Add mmal hardware video encoder support

lherman-cs:
Using mmal significantly boosts video fps by offloading the encoding work from CPU to GPU. On a Raspberry Pi 3, libx264 only gives 480p18, whereas mmal can give 720p30.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,15 +30,15 @@ jobs:
             libvpx-dev \
             libx264-dev
       - name: go vet
-        run: go vet ./...
+        run: go vet $(go list ./... | grep -v mmal)
       - name: go build
-        run: go build ./...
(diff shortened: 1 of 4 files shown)


at-wat:
Is it possible to build it on CI (maybe by `GOARCH=`)?

.github/workflows/ci.yaml
```diff
@@ -30,15 +30,15 @@ jobs:
             libvpx-dev \
             libx264-dev
       - name: go vet
-        run: go vet ./...
+        run: go vet `go list ./... | grep -v mmal`
```
at-wat, line 33: `$()` would be better than backquotes for safer nesting.

lherman-cs: Updated.

PR #241, opened 2020-10-26
Fix undiscovered some camera devices

lherman-cs:
/dev/v4l/by-path doesn't return all available devices. So, to make sure that we include all available devices, the list of devices will also complement with /dev/video*.


at-wat (approved):
LGTM

PR #242, opened 2020-10-26
Fix step wise resolutions in linux camera

lherman-cs:
Some cameras support a range of resolutions with step wise. The fix is to not only capture the highest resolutions but uses the step wise to determine if we can support the hardcoded standard resolutions, https://commons.wikimedia.org/wiki/File:Vector_Video_Standards2.svg. In the future, we should use a custom data structure to capture more resolutions that are outside of the listed standard resolutions.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -25,6 +25,36 @@ const (
 var (
 	errReadTimeout = errors.New("read timeout")
 	errEmptyFrame  = errors.New("empty frame")
+	// Reference: https://commons.wikimedia.org/wiki/File:Vector_Video_Standards2.svg
+	supportedResolutions = [][2]int{
+		{320, 240},
+		{640, 480},
+		{768, 576},
+		{800, 600},
+		{1024, 768},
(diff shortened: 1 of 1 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -23,8 +23,13 @@ const (
...
 
 var (
-	errReadTimeout = errors.New("read timeout")
-	errEmptyFrame  = errors.New("empty frame")
+	errReadTimeout       = errors.New("read timeout")
+	errEmptyFrame        = errors.New("empty frame")
+	supportedResolutions = [][2]int{
+		{1920, 1080}, // 1080p
+		{1280, 720},  // 720p
+		{640, 480},   // 480p
```
at-wat, line 31: Maybe better to have https://commons.wikimedia.org/wiki/File:Vector_Video_Standards2.svg these sizes?

lherman-cs: Yep, makes sense. Updated with these resolutions.

at-wat (approved):

PR #243, opened 2020-10-29
Update readers to be memory pool friendly

lherman-cs:


pkg/codec/codec.go
```diff
@@ -58,7 +58,7 @@ type VideoEncoderBuilder interface {
 
 // ReadCloser is an io.ReadCloser with methods for rate limiting: SetBitRate and ForceKeyFrame
 type ReadCloser interface {
-	Read() ([]byte, error)
+	Read() (b []byte, release func(), err error)
```
at-wat, line 61: @lherman-cs could you add a documentation how `release()` should be treated?
Will any implementation cause memory leak if `release()` is not called?

at-wat: Should it be called when `err` is non nil?

lherman-cs: @at-wat my bad, I should've documented it. I'll add this comment!

Will any implementation cause memory leak if release() is not called? **No, release() calls should be optional. The result is only going to be less optimal as it'll put a lot of pressure to the garbage collector, users might experience some freezes in the stream randomly. As of now, `sync.Pool` is the best candidate that I can think of to implement the memory pool, but this can change and it'll be transparent to the caller.**

Should it be called when err is non nil? **No, because the reader will not receive a buffer from the source**

at-wat: Thanks!

PR #244, opened 2020-10-30
New mediadevices design

lherman-cs:
Changelog:
  * Better support for non-webrtc use cases
  * Enable multiple readers
  * Enhance codec selectors
  * Update APIs to reflect on the new v3 webrtc design
  * Cleaner APIs

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -1,0 +1,117 @@
+package mediadevices
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/pion/mediadevices/pkg/codec"
+	"github.com/pion/mediadevices/pkg/io/audio"
+	"github.com/pion/mediadevices/pkg/io/video"
+	"github.com/pion/mediadevices/pkg/prop"
+	"github.com/pion/webrtc/v2"
+)
+
+// CodecSelector is a container of video and audio encoder builders, which later will be used
+// for codec matching.
+type CodecSelector struct {
+	videoEncoders []codec.VideoEncoderBuilder
+	audioEncoders []codec.AudioEncoderBuilder
+}
+
+// CodecSelectorOption is a type for specifying CodecSelector options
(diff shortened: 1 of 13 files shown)


lherman-cs:
@at-wat yes, the latest version doesn't support `rtp-send` use case yet, this is mainly to make the redesign work less daunting by breaking it apart to multiple small chunks (sorry, I should've notified you first, just really wanted to remove the friction 😓) . I also reorganized the examples so that they become more use-case based. I'm imagining to have the following examples:
* webrtc
* http
* rtp
* transformation

I already have a design in mind. I think the video track and audio track will have an API called `NewRTPReader`, which will create a new reader from broadcaster, build the encoder, and packetize the samples already. So, from the user's perspective it'll look like the following:

```go
codecSelector := mediadevices.NewCodecSelector(
  mediadevices.WithVideoEncoders(&vp8Params)
)
s, err := mediadevices.GetUserMedia(mediadevices.MediaStreamConstraints{
    c.FrameFormat = prop.FrameFormat(frame.FormatYUY2)
    c.Width = prop.Int(640)
    c.Height = prop.Int(480)
  },
  Codec: codecSelector,
})

t := s.GetVideoTracks()[0]
videoTrack := t.(*mediadevices.VideoTrack)

reader, err := videoTrack.NewRTPReader("vp8")
must(err)

client := newRTPClient()
buf := make([]byte, mtu)

for {
  // pkts = []*rtp.Packet
  pkts, release, err := reader.Read()
  for _, pkt := range pkts {
    pkt.MarshalTo(buf)
    rtpClient.Write(buf)
  }
  release()
}

```

lherman-cs:
@at-wat I just added the API that I mentioned earlier, and recreated the example with the new design, https://github.com/pion/mediadevices/blob/rtp/examples/rtp/main.go. 

What do you think about the new API?

at-wat:
@lherman-cs the API seems good!

PR #249, opened 2020-11-09
Fix wrong duration in vpx due to buffer usage

lherman-cs:
VPX doesn't allow 0 duration. If 0 is given, vpx_codec_encode will fail with VPX_CODEC_INVALID_PARAM.
0 duration is possible because mediadevices first gets the frame meta data by reading from the source,
and consequently the codec will read the first frame from the buffer. This means the first frame won't
have a pause to the second frame, which means if the delay is <1 ms (vpx duration resolution), duration
is going to be 0.

This is only a temporary fix for now. The main issue is tracked here, https://github.com/pion/mediadevices/issues/250.


at-wat:
Thanks, looks good to me

PR #258, opened 2020-11-23
Bring back x11 adapter for Linux build only

lherman-cs:
Original x11 adapter uses ~40% less CPU usage than kbinani/screenshot.
Ideally, we should migrate 100% to kbinani/screenshot. But, the CPU
usage difference is substantial. In the future, we should look deeper
into kbinani/screenshot and try to improve it.


at-wat (approved):
LGTM

PR #264, opened 2020-12-13
Enhance cross-platform build system

lherman-cs:
With this new build system, it'll make it easier to compile third-party libraries statically to multiple platforms.

PR #266, opened 2020-12-15
Enhance cross-platform build system

lherman-cs:
With this new build system, it'll make it easier to compile third-party libraries statically to multiple platforms.

Obsoletes https://github.com/pion/mediadevices/pull/264


lherman-cs:
> I think we should add pkg/codec/openh264/LICENSE file (copy of https://github.com/cisco/openh264/blob/master/LICENSE) to make the license information detectable by license discovery tool.

@at-wat what license discovery tool is this? I added the License, but I'm not familiar with the tool, and not sure how it'll be benefical.

at-wat:
I'm using https://github.com/google/go-licenses for Go.

lherman-cs:
> I'm using https://github.com/google/go-licenses for Go.

Neat! Thanks for sharing it

at-wat (approved):

PR #267, opened 2020-12-15
Add license check to CI

lherman-cs:
Add an automated license check to CI. Since we're using various third-party libraries, this automated check should help us in the long run.

Suggested by @at-wat: https://github.com/pion/mediadevices/pull/264#pullrequestreview-550878544

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,3 +81,17 @@ jobs:
         run: go test . pkg/... -v
         env:
           CGO_ENABLED: 0
+  check-licenses:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -81,3 +81,16 @@ jobs:
         run: go test . pkg/... -v
         env:
           CGO_ENABLED: 0
+  check-licenses:
+    name: Check Licenses
```
at-wat, line 85: https://github.com/pion/mediadevices/actions/runs/424104224
> The workflow is not valid. .github/workflows/ci.yaml (Line: 85, Col: 5): Required property is missing: runs-on

PR #268, opened 2020-12-15
Add static-linked opus option

lherman-cs:
Add static-linked opus option

Changes:
  * Remove hraban/opus dependency
  * Add multiple static-linked libopus binaries
  * Fix missing sampling rate in audiotest
  * Fix invalid format type panic


at-wat (commented):
License of the bundled opus library is https://github.com/xiph/opus/blob/master/COPYING. It should be copied under pkg/codec/opus.

lherman-cs:
> License of the bundled opus library is https://github.com/xiph/opus/blob/master/COPYING. It should be copied under pkg/codec/opus.

@at-wat this is the LICENSE that I'm using. Do you think I should rename it to COPYING instead?

at-wat:
Oh, I just overlooked the file. LICENSE should be fine.

at-wat (approved):

PR #269, opened 2020-12-16
Upgrade pion/webrtc from v2 to v3

lherman-cs:
With webrtc v3, users no longer need to bind or unbind manually anymore.


Sean-Der:
@lherman-cs Overall what do you think? Any changes we can/should make?

I am gonna spend the next few weeks upgrading random projects on Github to /v3

lherman-cs:
> @lherman-cs Overall what do you think? Any changes we can/should make?
> 
> I am gonna spend the next few weeks upgrading random projects on Github to /v3. Before I do that want to make sure these API changes were what we needed.

@Sean-Der I think overall the API is really good! I only have some minor comments:

1. Are we planning to have outbound MTU to be defined by the user? If yes, the users have to define this MTU in pion/webrtc SettingsEngine, this also means that`TrackLocalContext` should also expose MTU so that `mediadevices` can retrieve the configured MTU.
2. When there's an error in a video/audio stream, I think `TrackLocalContext` should have an API for the implementor to notify this error so that the PeerConnection can properly clean up.

PR #277, opened 2020-12-22
Add in a blocking ICE-Candidate function

wawesomeNOGUI:
#### Description
To be in spec with webrtc/v3 I replaced the setLocalDescription to a blocking non-trickle ICE function. I probably did it wrong though so feel free to change it!


examples/webrtc/main.go
```diff
@@ -99,24 +99,25 @@ func main() {
...
+	// Set the remote SessionDescription
+	err = peerConnection.SetRemoteDescription(offer)
 	if err != nil {
 		panic(err)
 	}
 
+
+
 	// Output the answer in base64 so we can paste it in browser
 	fmt.Println(signal.Encode(answer))
```
lherman-cs, line 122: This should be:

```go
<-webrtc.GatheringCompletePromise(peerConnection)
fmt.Println(signal.Encode(*peerConnection.LocalDescription()))
```

lherman-cs (changes requested):
This is a good catch!! Thanks! 

For the change itself, it's not necessary to change the sequence. We only need to add a wait for the ICE gathering to complete by using the helper `GatheringCompletePromise`, and get the updated SDP with the gathered candidates as the answer.

examples/webrtc/main.go
wawesomeNOGUI: Cool! I edited the code

examples/webrtc/main.go
```diff
@@ -99,25 +99,27 @@ func main() {
 		}
 	}
 
-	// Set the remote SessionDescription
```
lherman-cs, line 102: My bad, I think I was not clear 😅. These changes should not be needed, we still need to set remote description and local description to set the right state to PeerConnection and spin up the ICE gatherer.

examples/webrtc/main.go
```diff
@@ -99,25 +99,27 @@ func main() {
...
+	
+	// Set the remote SessionDescription
+	err = peerConnection.SetRemoteDescription(offer)
 	if err != nil {
 		panic(err)
 	}
 
-	// Output the answer in base64 so we can paste it in browser
-	fmt.Println(signal.Encode(answer))
+        //Wait for ICE gathering to complete (non-trickle ICE)
```
lherman-cs, line 115: The spaces seem to be off. Could you run `go fmt`?

lherman-cs (changes requested):

examples/webrtc/main.go
wawesomeNOGUI: Since we are only initiating createAnswer to start gathering ICE candidates (I think that’s right) I went ahead and set that equal to a blank identifier.

lherman-cs: The ICE gathering starts in SetLocalDescription. We need SetRemoteDescription to know whose side that's controlling.

examples/webrtc/main.go
wawesomeNOGUI: Ok I think my latest commit fixed this. Is it right that setRemoteDescription starts ICE gathering and uses the info set by setLocalDescription?

wawesomeNOGUI: Ohhh dear there's a lot to learnXD

wawesomeNOGUI: Ok so I think answer is a SDP that's supposed to be sent to the browser, but is the data at the memory location *peerConnection.LocalDescription() an SDP with updated ICE candidates? so we only have to send `fmt.Println(signal.Encode(*peerConnection.LocalDescription()))` for the user to paste in their browser?

examples/webrtc/main.go
wawesomeNOGUI: I ran the format on the latest commit

examples/webrtc/main.go
```diff
@@ -111,13 +111,17 @@ func main() {
...
 
-	// Sets the LocalDescription, and starts our UDP listeners
+	// Set the local SessionDescription
 	err = peerConnection.SetLocalDescription(answer)
 	if err != nil {
 		panic(err)
 	}
 
+	//Wait for ICE gathering to complete (non-trickle ICE)
+	<-webrtc.GatheringCompletePromise(peerConnection)
```
Sean-Der, line 121: You need to create the promise before `SetLocalDescription` if gathering finishes really quickly the resolve can happen before you create the promise (then it will just block forever)

* Create Promise
* SetLocalDescription
* Wait for promise

wawesomeNOGUI: Cool! So the ICE gatherer will update the SDP at `*peerConnection.LocalDescription()` multiple times and it's good to block until it finds the best candidate?

wawesomeNOGUI: Hmmm still getting the same error

Sean-Der: Yea! Looking at the code that all looks right. That matches what we do in all our other examples.

I am AFK for a bit but will try tonight!

wawesomeNOGUI: Cool thank you!

lherman-cs: @Sean-Der wouldn't it be fine since we have this check, https://github.com/pion/webrtc/blob/v3.0.0/gathering_complete_promise.go#L15?

wawesomeNOGUI:
Hi! I went ahead and rewrote the example to follow the Data-Channels-Create example in Pion/webrtc. So the go program creates the offer and the browser creates an answer. In theory it should work the opposite way too, but I couldn't figure out how to make it work that way.

examples/webrtc/main.go
lherman-cs: Yep, that's right!

lherman-cs:
@wawesomeNOGUI I'm not sure what happened. But, the diff is huge! I'm not sure what has changed.

examples/webrtc/main.go
wawesomeNOGUI: Cool!

wawesomeNOGUI:
:D Yeah, I changed a bunch, I'll add comments in the files changed section.

examples/webrtc/index.html
```diff
@@ -0,0 +1,76 @@
...
+        })
+
+
+        window.startSession = () => {
+          const sd = document.getElementById('remoteSessionDescription').value
+          if (sd === '') {
+            return alert('Session Description must not be empty')
+          }
+          try {
+            pc.setRemoteDescription(new RTCSessionDescription(JSON.parse(atob(sd)))).catch(log)
```
wawesomeNOGUI, line 66: Here I just copied how the data-channels-create is set up. Golang provides an offer, the browser gives an answer. After ` pc.createAnswer().then(d => pc.setLocalDescription(d)).catch(log)` is run, the browser starts gathering ice candidates and the final candidate gets pasted in the browser base64 description (line 47).

examples/webrtc/main.go
```diff
@@ -1,123 +1,133 @@
...
+                                Direction: webrtc.RTPTransceiverDirectionSendonly,
+                        },
+                )
+                if err != nil {
+                        panic(err)
+                }
+        }
+
+        // Create an offer
+        offer, err := peerConnection.CreateOffer(nil)
```
wawesomeNOGUI, line 102: Here I have the Golang program create an offer that the user pastes into the Golang base64 Session Description box in the html file. Then the user pastes the browser's answer into the golang console. At first I thought this was the only way to make the example work, because I only could get the example working by having the golang program offer and browser answer, but now I think I might've just made an error in my previous Html code. I'll investigate real quick.

wawesomeNOGUI:
Here's a [fiddle](https://jsfiddle.net/wawesome/8v97zc0u/1/)

Sean-Der:
Thank you so much for doing this! It has helped so many people in Slack :) 

I took your change in a359005a7d811b4f3f35b2acd996690a5958ce28 and it worked perfectly. Sorry this didn't get merged sooner. I really appreciate the work you are doing. If you are interested I have lots of other things that I need help on @wawesomeNOGUI that will be way quicker. 

If you are interested would love to add you to Pion org and throw some fun tasks your way!

wawesomeNOGUI:
Thank you so much, that sounds great! Working on Pion has been a great
experience and I would love to keep contributing!

PR #283, opened 2021-01-07
Fix VPX_CODEC_INVALID_PARAM on resolution change

f-fl0:
#### Description
The following error occurs when scaling up the video frames using transform function such as `video.Scale(width, height, video.ScalerFastBoxSampling)` with `width` or `height` greater than the original frame dimension:
```
vpx_codec_enc_config_set failed (8)
```

Although this block seems to work fine when scaling down the frame, it does not seem to be enough in this case:
```
if ec := C.vpx_codec_enc_config_set(e.codec, e.cfg); ec != C.VPX_CODEC_OK {
 	return 0, fmt.Errorf("vpx_codec_enc_config_set failed (%d)", ec)
}
```

Creating and initializing a new codec object seem necessary.

### DIFF
--- a/pkg/codec/vpx/vpx.go
+++ b/pkg/codec/vpx/vpx.go
@@ -229,9 +229,15 @@ func (e *encoder) Read() ([]byte, func(), error) {
 
 	if e.cfg.g_w != C.uint(width) || e.cfg.g_h != C.uint(height) {
 		e.cfg.g_w, e.cfg.g_h = C.uint(width), C.uint(height)
-		if ec := C.vpx_codec_enc_config_set(e.codec, e.cfg); ec != C.VPX_CODEC_OK {
-			return nil, func() {}, fmt.Errorf("vpx_codec_enc_config_set failed (%d)", ec)
+
+		newCodec := C.newCtx()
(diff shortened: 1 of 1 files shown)


pkg/codec/vpx/vpx.go
```diff
@@ -229,9 +229,15 @@ func (e *encoder) Read() ([]byte, func(), error) {
...
-		if ec := C.vpx_codec_enc_config_set(e.codec, e.cfg); ec != C.VPX_CODEC_OK {
-			return nil, func() {}, fmt.Errorf("vpx_codec_enc_config_set failed (%d)", ec)
+
+		newCodec := C.newCtx()
+		if ec := C.vpx_codec_enc_init_ver(
+			newCodec, e.codec.iface, e.cfg, 0, C.VPX_ENCODER_ABI_VERSION,
+		); ec != 0 {
+			return nil, func() {}, fmt.Errorf("vpx_codec_enc_init failed (%d)", ec)
 		}
+		e.codec = newCodec
```
at-wat, line 239: Old `e.codec` should be freed by `C.free(unsafe.Pointer(e.codec))`.

at-wat (approved):
LGTM

Issue #284, opened 2021-01-07
webrtc example fails on raspberry pi 3 due to opus linking error


andrein:
Version: d86fc4a3e92f120a390079974298902465533685
Hardware: Raspberry Pi 3 running an up to date Raspbian Buster

```
pi@raspberrypi:~/go/src/github.com/pion/mediadevices/examples $ make
cd archive && go build
cd facedetection && go build
cd http && go build
cd rtp && go build
cd webrtc && go build
# github.com/pion/mediadevices/pkg/codec/opus
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(opus_encoder.c.o): relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(repacketizer.c.o): relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(analysis.c.o): relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(mlp.c.o): relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(enc_API.c.o): relocation R_ARM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(encode_indices.c.o): relocation R_ARM_MOVW_ABS_NC against `silk_type_offset_VAD_iCDF' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: ../../pkg/codec/opus/lib/libopus-linux-armv7.a(encode_pulses.c.o): relocation R_ARM_MOVW_ABS_NC against `silk_max_pulses_table' can not be used when making a shared object; recompile with -fPIC
[...]

PR #286, opened 2021-01-08
Recompile Opus with PIC

lherman-cs:
Resolves https://github.com/pion/mediadevices/issues/284


andrein:
That seemed to do the trick, I was able to compile this branch and run the examples successfully. Thanks!

Issue #292, opened 2021-01-22
GetMediaDevice does not work on new Macbook Pro


rivercheng:
It worked on my old Macbook Pro, but failed on my new one. 
================ Best Fit ================

Not found
panic: failed to find the best driver that fits the constraints

rivercheng:
Seems to be the camera, because if I replace the 'camera' with 'videotest', then there's no error anymore. Maybe the new Macbook pro needs a new video driver?

rivercheng:
The new Macbook pro is MacBook Pro (16-inch, 2019)

lherman-cs:
@rivercheng thanks for reporting! This issue seems to be similar to https://github.com/pion/mediadevices/issues/287. Which version of `mediadevices` are you using? Could you try to pull the latest change which contains the fix, https://github.com/pion/mediadevices/issues/287? 

If that still doesn't fix it, could you try to run `ffmpeg -f avfoundation -framerate 30 -i default`, it should give us a list of supported formats by your camera. Very likely, we just need to add another mapping from AVFoundation to mediadevices.

nrwinner:
I can confirm I am also experiencing this issue with the same computer model. I'm using the most recent version of `mediadevices`. When I run that command, I see
```
[avfoundation @ 0x7f8051814c00] Selected pixel format (yuv420p) is not supported by the input device.
[avfoundation @ 0x7f8051814c00] Supported pixel formats:
[avfoundation @ 0x7f8051814c00]   uyvy422
[avfoundation @ 0x7f8051814c00]   yuyv422
[avfoundation @ 0x7f8051814c00]   nv12
[avfoundation @ 0x7f8051814c00]   0rgb
[avfoundation @ 0x7f8051814c00]   bgr0
```

daniel-d-truong:
^ just wanted to bump as i have the same issues as well

lherman-cs:
@nrwinner thanks for the extra information! From the output, `mediadevices` support both `uyvy422` and `yuyv422`. The FFmpeg mapping can be found here, https://github.com/FFmpeg/FFmpeg/blob/c810a9502cebe32e1dd08ee3d0d17053dde44aa9/libavdevice/avfoundation.m#L53-L80. The mapping is matched to what `mediadevices` has, https://github.com/pion/mediadevices/blob/master/pkg/avfoundation/AVFoundationBind/AVFoundationBind.m#L140-L145.

[...]

PR #295, opened 2021-01-28
Add formats and logging to AVFoundation backend

lherman-cs:
Changes:
* Add logging for unincluded raw frame formats
* Add NV12 support for avfoundation
* Fix wrong Cb and Cr order in NV21

Possibly resolves https://github.com/pion/mediadevices/issues/292


daniel-d-truong:
`GetMediaDevice` works for me now!

PR #301, opened 2021-02-17
Return selected codec information in NewEncodedReader and NewEncodedIOReader.

f-fl0:
#### Description

In this PR, a pointer to the selected codec is added to the results returned by `NewEncodedReader` and `NewEncodedIOReader`. At the moment only `newEncodedReader` returns this information but it is not accessible outside of the `track` class. The selected codec information might useful if ones wants to use a custom RTP `Packetizer`. This is also the reason why it needs to
return `rtp.Codec` instead of `webrtc.RTPCodecParameters`: `rtp.NewPacketizer` needs a `rtp.Payloader`.

### DIFF
--- a/examples/archive/main.go
+++ b/examples/archive/main.go
@@ -68,7 +68,7 @@ func main() {
 		})
 	}))
 
-	reader, err := videoTrack.NewEncodedIOReader(x264Params.RTPCodec().MimeType)
+	reader, _, err := videoTrack.NewEncodedIOReader(x264Params.RTPCodec().MimeType)
 	must(err)
 	defer reader.Close()
 
--- a/mediastream_test.go
+++ b/mediastream_test.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"testing"
 
+	"github.com/pion/mediadevices/pkg/codec"
 	"github.com/pion/webrtc/v3"
 )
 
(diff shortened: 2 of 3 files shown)


at-wat:
@lherman-cs could you take a look? I don't know very well about the intent of the original API design.

lherman-cs:
@f-fl0 the original idea was to take away the complexity of choosing the codec from the users but still giving the flexibility for RTP packetizer implementation. So, for writing your own Packetizer, you can achieve this by registering your own codec, e.g. https://github.com/pion/mediadevices/blob/master/pkg/codec/x264/params.go#L1-L50. Instead of returning the default RTP packetizer in RTPCodec, you can return the customized version.

at-wat:
It's bit off from this PR's topic, but in case that codec selector is like
```
mediadevices.NewCodecSelector(mediadevices.WithVideoEncoders(vaapiParam, vpxParam))
```
to use hardware accelerated encoder if available and fallback to libvpx, I think there should be a way to know which encoder was actually selected.
For this purpose, `codec.VideoEncoderBuilder` may be needed to be returned instead of `codec.RTPCodec`. (it requires bit more deeper changes.)

lherman-cs:
@at-wat that's a good point. I'm totally on board with the change. But, I'm curious, what's the use case of knowing between software vs hardware encodings?

at-wat:
> But, I'm curious, what's the use case of knowing between software vs hardware encodings?

I think notifying users that hardware encoder is used or not is helpful to validate resource usage.

f-fl0:
@lherman-cs @at-wat thank you for the feedback. I understand the modifications introduced here are not necessary. I will close this PR.

PR #304, opened 2021-02-28
prefer reallink for linux camera label

edaniels:
#### Description

Using the reallink base name allows one to actually filter for a device like `video0` when v4l reports the more complicated form. This also fixes `video0` being set as the high priority device if v4l finds it first.


at-wat (commented):
Correspondence of the hardware and `/dev/videoN` depends on the order of device recognition by linux and it is not fully stable when rebooting system with multiple cameras.
In my opinion, label should have a information that is invariable against timing.

What do you think?

edaniels:
So it can be stable based on udev rules in some systems. An alternative approach is to provide some more metadata from udev in the driver info (possibly an extra data field)?

at-wat:
Adding extra data field sounds bit difficult since it is based on the WebAPI MediaDeviceInfo structure (https://developer.mozilla.org/en-US/docs/Web/API/MediaDeviceInfo).

Another idea is to store both `/dev/video*` and `/dev/v4l/by-path/*` by joining them with separator character (or string) which is not appears in these paths.
e.g. `/dev/video0;/dev/v4l/by-path/pci-0000:00:00.0-usb-0:0:0.0-video-index0`

edaniels:
I like the semicolon. I’ll whip something up later today. I’ll also get on the other PR too. Thanks!

edaniels:
@at-wat Done. We just need to merge https://github.com/pion/mediadevices/pull/308 first.

edaniels:
whoops did not mean to close

edaniels:
Reopened at https://github.com/pion/mediadevices/pull/311

PR #307, opened 2021-03-01
Fix 32-bit usage of x11capture_linux

edaniels:
#### Description

This allows compilation on 32-bit devices.


at-wat (approved):
Good catch! Thank you

PR #313, opened 2021-03-04
Use correct NV12 decoder in camera_linux

edaniels:


at-wat:
This seems be from very beginning of the repository: https://github.com/pion/mediadevices/commit/5e797785dfe713eff328680f23e8db3def6d8b28#diff-9de2dbd5ae8591f0020859d31139841decfba8d061afa8b4e6ef65bc1640a44eR21

I'm suspecting that `decodeNV21` and `decodeNV12` are also swapped as no one have reported that color of captured image is wrong.

at-wat:
https://www.kernel.org/doc/html/v5.9/userspace-api/media/v4l/pixfmt-nv12.html
> V4L2_PIX_FMT_NV21 is the same except the Cb and Cr bytes are swapped, **the CrCb plane starts with a Cr byte.**

`decodeNV21`
https://github.com/pion/mediadevices/blob/eaaaacfc6b938bff60884dcf45867aad5ab10498/pkg/frame/yuv.go#L38-L39

Decoder name looks corresponding?

edaniels:
So it says "the CrCb plane starts with a Cr byte." for NV21. In that case, I think it is correct now.

edaniels:
To be clear, both the decoders are correct, but the mappings were wrong.

lherman-cs (approved):
LGTM. I didn't realize that the formats are swapped in the Linux adapter. I swapped the decoder implementation in https://github.com/pion/mediadevices/commit/5e797785dfe713eff328680f23e8db3def6d8b28#diff-9de2dbd5ae8591f0020859d31139841decfba8d061afa8b4e6ef65bc1640a44eR21 because it didn't follow the format according to the fourcc website, https://www.fourcc.org/pixel-format/yuv-nv21/

PR #315, opened 2021-03-07
Add decoder support for pixel format BGRA & ARGB

zyxar:
#### Description
This PR adds decoder support for pixel format `BGRA` & `ARGB`, and enables them in pkg `avfoundation`.


pkg/frame/rgb.go
```diff
@@ -0,0 +1,44 @@
...
+	size := 4 * width * height
+	if size > len(frame) {
+		return nil, func() {}, fmt.Errorf("frame length (%d) less than expected (%d)", len(frame), size)
+	}
+	r := image.Rect(0, 0, width, height)
+	for i := 0; i < size; i += 4 {
+		*(*uint32)(unsafe.Pointer(&frame[i])) = func(v uint32) uint32 {
+			return (v & 0xFF00FF00) | (v&0xFF)<<16 | (v&0xFF0000)>>16
+		}(*(*uint32)(unsafe.Pointer(&frame[i])))
+		//frame[i], frame[i+2] = frame[i+2], frame[i]
```
at-wat, line 20: It seems like a conversion from `ARGB` to `GRAB`. (`frame[i]` is `A` and `frame[i+2]` is `G` in the original ARGB data)
Maybe the implementations of `decodeARGB` and `decodeBGRA` are swapped?

pkg/frame/rgb.go
```diff
@@ -0,0 +1,44 @@
...
+
+func decodeBGRA(frame []byte, width, height int) (image.Image, func(), error) {
+	size := 4 * width * height
+	if size > len(frame) {
+		return nil, func() {}, fmt.Errorf("frame length (%d) less than expected (%d)", len(frame), size)
+	}
+	r := image.Rect(0, 0, width, height)
+	for i := 0; i < size; i += 4 {
+		*(*uint32)(unsafe.Pointer(&frame[i])) = bits.RotateLeft32(*(*uint32)(unsafe.Pointer(&frame[i])), -8)
+		//frame[i], frame[i+1], frame[i+2], frame[i+3] = frame[i+1], frame[i+2], frame[i+3], frame[i]
```
at-wat, line 37: `BGRA` to `GRAB` ?

pkg/frame/rgb_test.go
```diff
@@ -0,0 +1,48 @@
+package frame
+
+import (
+	"fmt"
+	"testing"
+)
+
```
at-wat, line 7: It would be nice to have a simple test like: https://github.com/pion/mediadevices/blob/dafd208de79bcf3308e4ea9e5724ad27f91c5051/pkg/frame/yuv_test.go#L10-L41

pkg/frame/rgb.go
zyxar: I think it only matters to the *name* of the pixel format.

according to libyuv [formats](https://chromium.googlesource.com/libyuv/libyuv/+show/master/docs/formats.md):
```c
kCMPixelFormat_32ARGB          = 32,      FOURCC_BGRA
kCMPixelFormat_32BGRA          = 'BGRA',  FOURCC_ARGB
```
and I use the *FOURCC_\** naming to to name `frame.FormatBGRA` and `frame.FormatARGB` which results a similar mapping in `avfoundation` pkg.

at-wat: Sorry, I misread endian. (It should be little endian on x86 and ARM)

This func seems converting 0xXXYYZZWW (input) to 0xXXWWZZYY (output).
As the output byte order for `image.RGBA` is RGBA order (XX=A, WW=B, ZZ=G, YY=R on little endian uint32), input byte order shoud be BGRA.

I guess
https://github.com/pion/mediadevices/pull/315/files#diff-99707109ce57457c8ab1161707d059373858d14ae66e06aed05c50501e29eaa6R148-R153
https://github.com/pion/mediadevices/pull/315/files#diff-99707109ce57457c8ab1161707d059373858d14ae66e06aed05c50501e29eaa6R179-R184
are reversed and name of `decodeARGB` and `decodeBGRA` should be swapped?

(I don't have any Macintosh computers, so I couldn't test it on the real environment.)

edaniels:
HI @zyxar, do you still want to get this in?

PR #311, opened 2021-03-02
combine reallink and label for cameras on linux

edaniels:
See https://github.com/pion/mediadevices/pull/304

PR #316, opened 2021-03-08
Fix linux camera label rule document

at-wat:
Fix https://github.com/pion/mediadevices/pull/311/files#r589148142


edaniels:
good catch! I'd approve but I can't :P. LGTM

at-wat:
Order of `video0` and `pci-0000:00:00.0-usb-0:0:0.0-video-index0` was also swapped.

at-wat:
@edaniels could you take another look? (added test to ensure the label naming rule)

edaniels:
Yep LGTM

PR #317, opened 2021-03-10
Fix audio codec latency handling

at-wat:
To avoid buffering audio data multiple times, remove buffer from
malgo audio driver and pass expected codec latency as a codec
parameter.

### DIFF
--- a/pkg/codec/opus/params.go
+++ b/pkg/codec/opus/params.go
@@ -1,27 +1,69 @@
 package opus
 
 import (
+	"time"
+
 	"github.com/pion/mediadevices/pkg/codec"
 	"github.com/pion/mediadevices/pkg/io/audio"
 	"github.com/pion/mediadevices/pkg/prop"
 	"github.com/pion/mediadevices/pkg/wave/mixer"
 )
 
+// Latency is a type of OPUS codec frame duration.
+type Latency time.Duration
+
+// Latency values available in OPUS codec.
+const (
+	Latency2500us Latency = Latency(2500 * time.Microsecond)
+	Latency5ms    Latency = Latency(5 * time.Millisecond)
+	Latency10ms   Latency = Latency(10 * time.Millisecond)
+	Latency20ms   Latency = Latency(20 * time.Millisecond)
+	Latency40ms   Latency = Latency(40 * time.Millisecond)
+	Latency60ms   Latency = Latency(60 * time.Millisecond)
+)
+
+// Validate that the Latency is allowed in OPUS.
+func (l Latency) Validate() bool {
+	switch l {
+	case Latency2500us, Latency5ms, Latency10ms, Latency20ms, Latency40ms, Latency60ms:
+		return true
+	default:
+		return false
+	}
+}
+
+// Duration returns latency in time.Duration.
+func (l Latency) Duration() time.Duration {
+	return time.Duration(l)
+}
+
+// samples returns number of samples for given sample rate.
+func (l Latency) samples(sampleRate int) int {
+	return int(l.Duration() * time.Duration(sampleRate) / time.Second)
+}
+
 // Params stores opus specific encoding parameters.
 type Params struct {
 	codec.BaseParams
 	// ChannelMixer is a mixer to be used if number of given and expected channels differ.
 	ChannelMixer mixer.ChannelMixer
+
+	// Expected latency of the codec.
+	Latency Latency
 }
 
 // NewParams returns default opus codec specific parameters.
 func NewParams() (Params, error) {
-	return Params{}, nil
+	return Params{
+		Latency: Latency20ms,
+	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/driver/microphone/microphone.go
```diff
@@ -155,9 +155,6 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
 		return decodedChunk, func() {}, err
 	})
 
-	// FIXME: The current audio detection and audio encoder can only work with a static latency. Since the latency from the driver
-	//        can fluctuate, we need to stabilize it. Maybe there's a better way for doing this?
-	reader = audio.NewBuffer(int(inputProp.Latency.Seconds() * float64(inputProp.SampleRate)))(reader)
```
at-wat, line 160: I think codec should have a buffer with length which is needed by the specific codec and the length of the buffer should be independent from audio driver implementation.

lherman-cs (commented):
LGTM. Left a minor comment for the API.

pkg/codec/opus/params.go
```diff
@@ -16,12 +18,18 @@ type Params struct {
 
 // NewParams returns default opus codec specific parameters.
 func NewParams() (Params, error) {
-	return Params{}, nil
+	return Params{
+		BaseParams: codec.BaseParams{
+			Latency: 20 * time.Millisecond,
```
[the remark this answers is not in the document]
at-wat: There may be two options:
- Put `Latency time.Duration` to `codec.BaseParams`
  - *pro:* provides a general way to pass expected latency to codecs. Other codecs may also have latency parameter. (e.g. vpx has a option to add some lag to achieve better quality and accurate bit rate control.)
  - *con:* can't disallow values which are not supported by the codec.
- Put `Latency CustomLatencyType` (`type CustomLatencyType time.Duration`) to `opus.Param`
  - *pro:* can restrict value according to the codec spec.
    - however, `CustomLatencyType` can store a value which is not defined as consts. Validation or nearest value select is still needed.
  - *con:* way to set latency will be different for each codec.

What do you think?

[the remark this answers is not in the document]
at-wat: (I don't have strong opinion about this.)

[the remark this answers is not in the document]
lherman-cs: I prefer using `CustomLatencyType`, because this allows users to know available latencies for specific codecs since the options will show up on Godocs. Whereas if we used a generic `time.Duration`, there's really no way to know the right values until the user reads our codebase and understands the specific codec. 

And, yes, I agree that we still have to check the actual value since Go doesn't have a way to enforce it at compile time.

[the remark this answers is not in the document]
at-wat: Added `Latency` type.

at-wat:
@lherman-cs please take another look when you have time

at-wat:
rebased and squashed

pkg/codec/opus/params.go
```diff
@@ -1,27 +1,69 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+// Duration returns latency in time.Duration.
+func (l Latency) Duration() time.Duration {
+	return time.Duration(l)
+}
```
lherman-cs, line 38: I'm not sure why this is needed as time.Duration can do this as well?

at-wat: Alias type requires explicit cast when calculating with the original type.

https://play.golang.org/p/jG0iPXAArA3

pkg/codec/opus/params.go
```diff
@@ -1,27 +1,69 @@
...
+	return Params{
+		Latency: Latency20ms,
+	}, nil
 }
 
 // RTPCodec represents the codec metadata
 func (p *Params) RTPCodec() *codec.RTPCodec {
-	return codec.NewRTPOpusCodec(48000)
+	c := codec.NewRTPOpusCodec(48000)
+	c.Latency = time.Duration(p.Latency)
```
at-wat, line 65:
```suggestion
	c.Latency = p.Latency.Duration()
```

at-wat:
rebased

pkg/codec/opus/params.go
at-wat: @lherman-cs what do you think?

at-wat:
@lherman-cs PTAL when you have time

pkg/codec/opus/params.go
lherman-cs: To me, an explicit cast is more idiomatic, for example casting a number to `time.Duration`. I'm okay with both approaches.

lherman-cs (approved):
LGTM

lherman-cs:
> @lherman-cs PTAL when you have time

Sorry for the delay 😓

at-wat:
Thank you!

Issue #325, opened 2021-03-31
Compile error in driver > wrapper.go


Junaid-Ahm:
This statement `uuid.NewV4().String()`  [here](https://github.com/pion/mediadevices/blob/master/pkg/driver/wrapper.go#L11) is buggy. 

The function `NewV4()` returns two values, string, error. Instead it's trying to use only the first value.

Junaid-Ahm:
Seems to be versioning issue with go-uuid package

lherman-cs:
@Junaid-Ahm I can't reproduce this on my machine. But, https://github.com/pion/mediadevices/pull/329 hopefully fixes this.

PR #329, opened 2021-04-11
Remove github.com/satori/go.uuid dependency

lherman-cs:
Resolves: https://github.com/pion/mediadevices/issues/325


pkg/driver/wrapper.go
```diff
@@ -1,14 +1,19 @@
...
+	"github.com/google/uuid"
 	"github.com/pion/mediadevices/pkg/io/audio"
 	"github.com/pion/mediadevices/pkg/io/video"
 	"github.com/pion/mediadevices/pkg/prop"
-	uuid "github.com/satori/go.uuid"
 )
 
 func wrapAdapter(a Adapter, info Info) Driver {
-	id := uuid.NewV4().String()
+	generator, err := uuid.NewRandom()
```
at-wat, line 11: `uuid.NewRandom()` returns binary UUID data and it's not generator. It would be better to be renamed.

at-wat (approved):
Commented a naming issue. The change looks good to me!
I have been using `github.com/google/uuid` in other projects for a long while and met no problem.

PR #334, opened 2021-05-10
Implements RequireKeyFrame for VPX codec

EmrysMyrddin:
#### Description

This PR implements Key Frame requests for VPX.

My goal is to use this when we receive a PLI packet to send a fresh new Key Frame.

### DIFF
--- a/pkg/codec/vpx/vpx_test.go
+++ b/pkg/codec/vpx/vpx_test.go
@@ -20,6 +20,8 @@ func TestImageSizeChange(t *testing.T) {
 		},
 		"VP9": func() (codec.VideoEncoderBuilder, error) {
 			p, err := NewVP9Params()
+			// Disable latency to ease test and begin to receive packets for each input frame
+			p.LagInFrames = 0
 			return &p, err
 		},
 	} {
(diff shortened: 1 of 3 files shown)


pkg/codec/vpx/vpx_test.go
```diff
@@ -87,3 +89,72 @@ func TestImageSizeChange(t *testing.T) {
...
+func TestRequestKeyFrame(t *testing.T) {
+	for name, factory := range map[string]func() (codec.VideoEncoderBuilder, error){
+		"VP8": func() (codec.VideoEncoderBuilder, error) {
+			p, err := NewVP8Params()
+			return &p, err
+		},
+		"VP9": func() (codec.VideoEncoderBuilder, error) {
+			p, err := NewVP9Params()
+			// Disable latency to ease test and begin to receive packets for each input frame
+			p.LagInFrames = 0
```
lherman-cs, line 102: From my understanding, LagInFrames defaults to 0, similar to VP8. Why VP9 need to be configured differently?

lherman-cs (commented):
Overall looks good to me! Only a minor comment.

pkg/codec/vpx/vpx_test.go
EmrysMyrddin: It seems that 0 is not the default for VP9 (I don't remember where I found this, I will update if I find it). This leaded in test to get empty packets.
For the first tests, it was not a big deal because we didn't care about the content of the packets.

For me, I need the packets to be sent directly for being able to check if I really get a key frame when requested.

lherman-cs (approved):

PR #337, opened 2021-05-19
Fix dead lock when closing peer connection with running tracks

### LINKED ISSUE #336: Dead lock on closing Peer Connection while media device track running
### Your environment.
  - Version: v0.2.0
  - Browser: none

### What did you do?
I have started a peer connection with a media device track sending camera video feed.
Call `Close` method on the peer connection.

### What did you expect?
Calling `Close` should close the connection as expected

### What happened?
Calling `Close` causes a dead lock.

The reason of the dead lock is a double lock request on the same mutex by both the track closing code and the track error handling code.

EmrysMyrddin:
#### Description
The `unbind` function was borrowing the track mutex more than it should. We need to borrow only while searching for active peer connection.

Keeping the lock while waiting for the end of the peer connection leads to dead lock with other parts of the code such as track's error handling code.

#### Reference issue
Fixes #336

### DIFF
--- a/track.go
+++ b/track.go
@@ -209,19 +209,28 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
 }
 
 func (track *baseTrack) unbind(ctx webrtc.TrackLocalContext) error {
-	track.mu.Lock()
-	defer track.mu.Unlock()
-
-	ch, ok := track.activePeerConnections[ctx.ID()]
-	if !ok {
-		return errNotFoundPeerConnection
+	ch, err := track.removeActivePeerConnection(ctx.ID())
+	if err != err {
+		return err
 	}
 
 	doneCh := make(chan struct{})
 	ch <- doneCh
 	<-doneCh
-	delete(track.activePeerConnections, ctx.ID())
 	return nil
+}
+
+func (track *baseTrack) removeActivePeerConnection(id string) (chan<- chan<- struct{}, error) {
+	track.mu.Lock()
+	defer track.mu.Unlock()
+
+	ch, ok := track.activePeerConnections[id]
+	if !ok {
+		return nil, errNotFoundPeerConnection
+	}
+	delete(track.activePeerConnections, id)
+
+	return ch, nil
 }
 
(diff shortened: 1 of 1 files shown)


lherman-cs (approved):
Good catch, thanks!!

track.go
```diff
@@ -209,21 +209,30 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
...
-	ch, ok := track.activePeerConnections[ctx.ID()]
-	if !ok {
-		return errNotFoundPeerConnection
+	ch, err := track.removeActivePeerConnection(ctx.ID())
+	if err != err {
+		return err
 	}
 
 	doneCh := make(chan struct{})
 	ch <- doneCh
```
itzmanish, line 218: Just a small question why are you doing this?
If you want to close a channel then you can do close(ch) instead.

track.go
```diff
@@ -209,21 +209,30 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
...
 	<-doneCh
```
itzmanish, line 219: I don't understand the use of doneCh and getting from empty doneCh.

track.go
at-wat: https://github.com/pion/mediadevices/blob/b3c94a1f7bc2a1ef43c4056258269083a7e760f5/track.go#L174-L192
`doneCh` is sent to a routine created by `bind()` through `ch` and the routine closes the `doneCh` after finishing the loop.
`<-doneCh` waits until finishing the loop.

itzmanish: Yep. Makes sense.

track.go
itzmanish: I am finding it difficult to understand that won't line 221 creates an error like `fatal error: all goroutines are asleep - deadlock!`? And line 222 will wait for forever since there is nothing in doneCh which will create deadlock situation.

itzmanish: https://play.golang.org/p/QufENXCZpO-

at-wat: https://play.golang.org/p/qBdpOKwdorR
as written in https://github.com/pion/mediadevices/pull/337#discussion_r739964814

itzmanish: Nice. This is a cool trick. Thanks @at-wat , lesson learned. 😃

PR #338, opened 2021-05-19
Force key frame on PLI

EmrysMyrddin:
#### Description
This is a first POC around handling PLI for codecs on which we have control over key frame generation.

I have added the idea of an `EncoderController` instead of adding `ForceKeyFrame` and `SetBitRate` directly on the `ReadCloser` interface.

This allows more flexibility on encoder implementation. The main benefit is that you can optionally implement bit rate and key frame control. In our case, only 2 codecs have this implementations.

Using interfaces for each controls allows to check if an encoder have this controls and enable features only on compatible encoders.

The next step probably to add handling of bit rate negotiation packets, but this PR is already pretty long so I don't want to put too much stuff in it for now.

#### Reference issue
This PR needs some modification in the `pion/webrtc` package for being able to read RTCP feedbacks. Here is the related PR that needs to be merged before this one: https://github.com/pion/webrtc/pull/1812


pkg/codec/codec.go
```diff
@@ -112,15 +112,37 @@ type VideoEncoderBuilder interface {
...
-// ReadCloser is an io.ReadCloser with methods for rate limiting: SetBitRate and ForceKeyFrame
+// ReadCloser is an io.ReadCloser with a controller
 type ReadCloser interface {
 	Read() (b []byte, release func(), err error)
 	Close() error
+	Controllable
+}
+
+// EncoderController is the interface allowing to control the encoder behaviour after it's initialisation
+// It will will possibly have common control method in the future.
```
at-wat, line 123:
```suggestion
// EncoderController is the interface allowing to control the encoder behaviour after it's initialisation.
// It will possibly have common control method in the future.
```

pkg/codec/mmal/mmal.go
```diff
@@ -91,12 +91,14 @@ func (e *encoder) Read() ([]byte, func(), error) {
...
 }
 
-func (e *encoder) SetBitRate(b int) error {
-	panic("SetBitRate is not implemented")
-}
+// TODO: Implement Bit rate control
+// var _ codec.BitRateController = (*encoder)(nil)
+
+// TODO: Implement Key Frame control
+//var _ codec.KeyFrameController = (*encoder)(nil)
```
at-wat, line 98: It may be nice to be in `_test.go` file to clarify it is interface type check.

pkg/codec/vpx/vpx.go
```diff
@@ -283,14 +283,19 @@ func (e *encoder) Read() ([]byte, func(), error) {
...
-	panic("SetBitRate is not implemented")
-}
+// TODO: Implement bit rate control
+//var _ codec.BitRateController = (*encoderVP8)(nil)
+
+var _ codec.KeyFrameController = (*encoder)(nil)
 
 func (e *encoder) ForceKeyFrame() error {
 	panic("ForceKeyFrame is not implemented")
 }
```
at-wat, line 293: This should be removed for now.

at-wat (commented):
API design looks nice to me. Left some minor comments.

pkg/codec/mmal/mmal.go
EmrysMyrddin: You mean moving the type check lines in a `_test.go` ?

at-wat: Yes, I think it's better to separate testing purpose code which is not used in the product into `*_test.go` files.

EmrysMyrddin: Ok, I have written real test instead of the hacky type check.

I think it is in fact better since it test the returned value of `Controller` is implementing the interfaces instead of hardcoded `encoder` type. This way, I don't make any assumption on how this is implemented.

EmrysMyrddin:
Stil waiting for https://github.com/pion/webrtc/pull/1812 to be merged for being able to make this PR ready to merge.

Do you have any insigh on what I can do to make it happen ?

.idea/misc.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Sean-Der, line 1: @EmrysMyrddin Remove this file please

PR #381, opened 2022-03-05
Fixing Stride Handling for OpenH264

adamroach:
The golang image.YCbCr struct can contain fields of arbitrary stride
lengths, which don't necessarily match the field widths. This
patch passes the stride values from the image.YCbCr struct into
the OpenH264 encode calls.


Sean-Der (approved):

at-wat (approved):
LGTM

Issue #312, opened 2021-03-03
ScalerFastBoxSampling produces incorrect results


f-fl0:
### Your environment.
  - Version: *0.1.17*

### What did you do?

```
package main

import (
	"fmt"
	"image"
	"image/jpeg"
	"os"

	"github.com/pion/mediadevices/pkg/io/video"
)

func main() {

	inFile, _ := os.Open("image.jpg")
	defer inFile.Close()

	img, _, _ := image.Decode(inFile)
	src := video.ReaderFunc(func() (image.Image, func(), error) {
		return img, func() {}, nil
	})

	scaleFBS := video.Scale(640, 480, video.ScalerFastBoxSampling)(src)
	scaled, _, _ := scaleFBS.Read()

	outFile, _ := os.Create(fmt.Sprintf("scaled.jpg"))
	defer outFile.Close()
	jpeg.Encode(outFile, scaled, nil)
}
```
where `image.jpg` is https://upload.wikimedia.org/wikipedia/commons/d/dc/DenglerSW-Parliament-Ottawa-20050730-1280x960.jpg
	
### What did you expect?
![scaled](https://user-images.githubusercontent.com/66578286/109743469-19dc7300-7c14-11eb-91b9-5b08cfbd61bb.jpg)

### What happened?
![scaled](https://user-images.githubusercontent.com/66578286/109743302-cff38d00-7c13-11eb-89c1-898ddd7fa692.jpg)

f-fl0:
Changing `ScalerFastBoxSampling` to `ScalerFastNearestNeighbor`, produces this image:
![scaled](https://user-images.githubusercontent.com/66578286/109753197-2c12dd00-7c25-11eb-8aad-e6ef7ee98540.jpg)

Using `ScalerFastBoxSampling` while removing the `if` condition from these lines https://github.com/pion/mediadevices/blob/b5b065369714d19004e7aedf18e8a186586b44a4/pkg/io/video/scaler_cgo.go#L158-L160 produces the same image.

at-wat:
What is the input [YCbCrSubsampleRatio](https://golang.org/pkg/image/#YCbCrSubsampleRatio)?

f-fl0:
> What is the input [YCbCrSubsampleRatio](https://golang.org/pkg/image/#YCbCrSubsampleRatio)?

```
	imgYCbCr := img.(*image.YCbCr)
	fmt.Println(imgYCbCr.SubsampleRatio)
[...]

PR #385, opened 2022-03-15
Fix scaling functions

f-fl0:
#### Description
This PR corrects two small errors that causes the improper results reported in the referenced issue:
- A copy paste typo (most likely) which sets the `Dy` component of the image rectangle to `Dx`.
- The `tmp` array in `fastBoxSampling` was not properly reset because the last argument in `memset` should be the number of bytes to fill in and not the number of elements.

#### Reference issue
Fixes https://github.com/pion/mediadevices/issues/312


at-wat (commented):
Could you make the existing test fails on master branch and success on this branch?

f-fl0:
I found another problem when using  the nearest neighbor scaling function when the same size as the input image:
image
 It is demonstrated by the test added in d56d4fb42127acf4b34b821c81aca55dc453f99b.

f-fl0:
The need for the fix introduced in ee67dee4b812bb68e026cfb0b57115d9fb8a1b45 is demonstrated by the test added in 5f2216ace1313f2e521e9fe4b5ee0b82bff0f590.

f-fl0:
The need for the fix introduced in ad6b0ad5e1c7878befe0dca92d82e09b3dae345d is demonstrated by the test added in edef5f5dee11be34eb8c12b719064b2764b6fd4d

f-fl0:
@at-wat I added some tests and tried to demonstrate they would fail on master by reverting each change and let CI fail.

pkg/io/video/scale_test.go
```diff
@@ -247,6 +525,103 @@ func TestScale(t *testing.T) {
...
+					0xD0, 0x98, 0x80,
+				},
+				YStride: 6,
+				CStride: 3,
+				Rect:    image.Rect(0, 0, 6, 4),
+			},
+		},
+	}
+
+	t.Run("FastBoxSampling", func(t *testing.T) {
```
at-wat, line 589: This Run seems redundant (same as `func TestScaleFastBoxSampling(t *testing.T)`)

pkg/io/video/scale_test.go
```diff
@@ -247,6 +525,103 @@ func TestScale(t *testing.T) {
 	}
 }
 
+func TestScaleFastBoxSampling(t *testing.T) {
+
```
at-wat, line 529:
```suggestion
```

at-wat (commented):
Overall looks good to me and added a few minor comments

pkg/io/video/scale_test.go
f-fl0: Removed in 466d14b1737a701bc31500700f8ef6f51147d76b.

pkg/io/video/scale_test.go
f-fl0: Removed in 466d14b1737a701bc31500700f8ef6f51147d76b.

at-wat (approved):
LGTM

PR #390, opened 2022-03-30
Update webcam module to set the camera framerate

f-fl0:
#### Description

This PR updates the [webcam](https://github.com/blackjack/webcam ) module to `ba064708e1652773bdb71e0605a1819ef4fad2e7` which allows to set (and get) the video camera capture framerate based on the `FrameRate` value specified in `prop.Media`.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -179,9 +179,11 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 		return nil, err
 	}
 
-	err = c.cam.SetFramerate(float32(p.FrameRate))
-	if err != nil {
-		return nil, err
+	if p.FrameRate > 0 {
+		err = c.cam.SetFramerate(float32(p.FrameRate))
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	if err := c.cam.StartStreaming(); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -179,6 +179,11 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 		return nil, err
 	}
 
+	err = c.cam.SetFramerate(float32(p.FrameRate))
```
at-wat, line 182: How it behaves when user doesn't explicitly specify frame rate?

f-fl0: Good point. I will check that.

f-fl0:
`15973f6961cbe56d44372d74f46abe88598265a1` failed in CI because of the following error:
```
=== RUN   TestThrottle
    throttle_test.go:36: Number of pushed images is expected to be 20, but pushed 27
--- FAIL: TestThrottle (0.54s)
FAIL
```

It seems irrelevant to the changes introduced in this PR.

pkg/driver/camera/camera_linux.go
f-fl0: I added some tests. When the framerate is not specified the constraint seems to be ignored.

at-wat: `p.FrameRate` seems to be 0 when user doesn't set frame rate.
I think it's better not to call SetFramerate in this case.

at-wat (approved):
LGTM

PR #394, opened 2022-04-05
Remove video framerate as explicit constraint

f-fl0:
#### Description
This PR addresses an issue introduced by changes from https://github.com/pion/mediadevices/pull/390: the video framerate was added as an explicit constraint to find a compatible driver. As `camera.Properties` currently does not have access to the list of available framerate no driver can be matched. Until this information can be retrieved it is better to not include the framerate in the fitness function. As a workaround, when a framerate is specified (i.e. > 0), a comparison between the desired value and the one reported by the device is performed and an error is returned when they are not equal.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -6,7 +6,6 @@ import "C"
 import (
 	"context"
 	"errors"
-	"fmt"
 	"image"
 	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -184,6 +185,16 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		framerate, err := c.cam.GetFramerate()
+		if err != nil {
+			return nil, err
+		}
+
+		if framerate != p.FrameRate {
```
at-wat, line 194: Framerate in float may have quantization error.
For example, there are 29.97 fps which can't be represented using IEEE754 float.
(This is why the internal v4l API uses fraction)

I'm not very sure we should return error if actual framerate is not equal to the desired value.
Cameras I have seems to select nearest available framerate if given value doesn't match any of available values.
In some case, user just want nearest framerate instead of the exact one. (on many USB cameras, framerate depends on the brightness of the image)

f-fl0: I can take this check out. I think it would be good to at least print something to the console to warn the user if the specified framerate is different that the one reported by the device.

f-fl0: Removed in 782d02e00c401f42e8cc0c3ca256ced1f5943afd.

at-wat (approved):
LGTM

PR #395, opened 2022-04-06
fix : Turn on the microphone with the current device id

zjzhang-cn:
#### Description
根据DeviceID打开mic，修复多个mic设备的不能选择的bug。
Open microphone with DeviceID. fix the bug that multiple mic devices cannot be selected.
#### Reference issue
Fixes #...

### DIFF
--- a/pkg/driver/microphone/microphone.go
+++ b/pkg/driver/microphone/microphone.go
@@ -111,6 +111,8 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
 	config.PerformanceProfile = malgo.LowLatency
 	config.Capture.Channels = uint32(inputProp.ChannelCount)
 	config.SampleRate = uint32(inputProp.SampleRate)
+	//FIXME: Turn on the microphone with the current device id
+	config.Capture.DeviceID = m.ID.Pointer()
 	if inputProp.SampleSize == 4 && inputProp.IsFloat {
 		config.Capture.Format = malgo.FormatF32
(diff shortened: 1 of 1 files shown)


pkg/driver/microphone/microphone.go
```diff
@@ -111,6 +111,8 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
 	config.PerformanceProfile = malgo.LowLatency
 	config.Capture.Channels = uint32(inputProp.ChannelCount)
 	config.SampleRate = uint32(inputProp.SampleRate)
+	//FIXME: Turn on the microphone with the current device id
```
at-wat, line 114: Is it not fixed yet and to be fixed in future? (FIXME?)

zjzhang-cn: It has been modified，

pkg/driver/microphone/microphone.go
```diff
@@ -111,6 +111,8 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
 	config.PerformanceProfile = malgo.LowLatency
 	config.Capture.Channels = uint32(inputProp.ChannelCount)
 	config.SampleRate = uint32(inputProp.SampleRate)
+	//FIX: Turn on the microphone with the current device id
```
at-wat, line 114: As it seems already fixed by this PR, removing `FIX:` would be less confusing.

Maybe
```suggestion
	// Select the microphone with the current device id
```
?

at-wat (approved):
The fix looks good to me

PR #405, opened 2022-05-13
Allow to enable frame copy on a video track

EmrysMyrddin:
#### Description

This PR aims to give access to copy frame feature of `Broadcaster` in high level functions such as `track.NewEncodedReader()`.

Today, `track.New*Reader()` functions forces to disable frames copy from Broadcaster. This is not convenient when you want to use high level API with multiple readers, leading to video artefacts on some codec/frame format combination.

This API is not very elegant but doesn't introduce any complex or breaking changes.


maeln:
Will this PR be merged ? If not what is holding it from being merged ?

Sean-Der (approved):

PR #406, opened 2022-06-01
Use wanted RTP payload type

matt-allan:
#### Description

This PR updates the track implementation to use payload type ID requested by the client.

Without this fix when you run the webrtc example with a non Chromium browser, the client requests a payload type ID such as 126 and we reply with a different type ID, such as 125. When this happens the stream connects but there's no output. Any payload type ID in the range 96-127 is dynamic so we have to negotiate it.

This does fix my issue buy I''m not really sure if this is the best way to fix this. Do we need to support non dynamic payload types too? Do we need to validate bounds or anything?

#### Reference issue

Fixes this issue I accidentally opened on pion/webrtc repo: https://github.com/pion/webrtc/issues/2237

### DIFF
--- a/track.go
+++ b/track.go
@@ -68,7 +68,7 @@ type Track interface {
 	// the encoded data in RTP format with given mtu size.
 	//
 	// Note: `mtu int` will be changed to `mtu uint16` in a future update.
-	NewRTPReader(codecName string, ssrc uint32, mtu int) (RTPReadCloser, error)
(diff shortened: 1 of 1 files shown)


EmrysMyrddin:
Hey ! Thank you for your proposal ! Extending compatibility by handling dynamic type ID would be great :-)

But i think we should find a way to do it transparently and automatically. And it would be awesome to avoid breaking the public API by doing it :-)

PR #408, opened 2022-06-17
Camera timeout duration as parameter

f-fl0:
#### Description

This PR adds a field to the `Video` and `VideoConstraints` to be able to set the timeout duration when reading frames from the video source. This field is not used in `fitnessDistance` and is only considered when set to a positive value. These changes should not modify the original behavior as the default value is kept to 5 seconds.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -192,6 +192,23 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 
 	cam := c.cam
 
+	// initialization
+	err = cam.WaitForFrame(5) // 5 seconds
+	switch err.(type) {
+	case nil:
+	case *webcam.Timeout:
+		return nil, errReadTimeout
+	default:
+		// Camera has been stopped.
+		return nil, err
+	}
+
+	// default to 5 seconds
+	var readTimeoutSec uint32 = 5
+	if p.ReadTimeoutSec > 0 {
+		readTimeoutSec = uint32(p.ReadTimeoutSec)
+	}
+
 	ctx, cancel := context.WithCancel(context.Background())
 	c.cancel = cancel
 	var buf []byte
@@ -207,7 +224,7 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 				return nil, func() {}, io.EOF
 			}
 
-			err := cam.WaitForFrame(5) // 5 seconds
+			err := cam.WaitForFrame(readTimeoutSec)
 			switch err.(type) {
 			case nil:
 			case *webcam.Timeout:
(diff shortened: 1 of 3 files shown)


pkg/prop/prop.go
```diff
@@ -229,16 +229,18 @@ func (c *comparisons) fitnessDistance() (float64, bool) {
 
 // VideoConstraints represents a video's constraints
 type VideoConstraints struct {
-	Width, Height IntConstraint
-	FrameRate     FloatConstraint
-	FrameFormat   FrameFormatConstraint
+	Width, Height  IntConstraint
+	FrameRate      FloatConstraint
+	FrameFormat    FrameFormatConstraint
+	ReadTimeoutSec IntConstraint
```
at-wat, line 235: As it is a driver specific config rather than a device selection constraint, I think it should be only in `pkg/driver/camera/`.
Possible implementation would be
- add package global variable to be configured (simple but might cause problem when set by multiple places)
- add package global unexported variable and set by compile option (by `-ldflags '-X module.import.path.variable=value'`)

at-wat: another possible implementation is:
- get value from an environment variable

f-fl0: I implemented a function in `camera_linux` to read the value from the `CAMERA_READ_TIMEOUT` environment variable if it exists and is appropriate.

pkg/driver/camera/camera_linux.go
```diff
@@ -133,6 +134,19 @@ func newCamera(path string) *camera {
 	return c
 }
 
+func getCameraReadTimeout() uint32 {
+	// default to 5 seconds
+	var readTimeoutSec uint32 = 5
+	if val, ok := os.LookupEnv("CAMERA_READ_TIMEOUT"); ok {
```
at-wat, line 140: It might be better to prefix something like  `PION_`/`PION_MEDIADEVICES_`

pkg/driver/camera/camera_linux.go
```diff
@@ -192,6 +206,16 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
...
 	cam := c.cam
 
+	// wait up to 5 seconds to read the first frame
+	if err := cam.WaitForFrame(5); err != nil {
+		if _, ok := err.(*webcam.Timeout); ok {
+			return nil, errReadTimeout
+		}
+		// Camera has been stopped.
+		return nil, err
+	}
```
at-wat, line 216: What is this aiming for?

Originally, user can read a new frame even after `video.Reader` returned an error, but this change will prevent to try reading after error.

f-fl0: The goal was to allow for a longer timeout read error at the beginning. I observed the device I'm using can take a few seconds to read the first frame. I will take this part out.

f-fl0: Removed in a8e90965b131100c9202c147c6633bdb7e35bb85

PR #414, opened 2022-07-25
Fix closing a camera not closing the associated track

### LINKED ISSUE #379: "panic: send on closed channel" after call Close on peerConnection
### Your environment.
  - Version: github.com/pion/mediadevices v0.3.2, github.com/pion/webrtc/v3 v3.1.10
  - Browser: Chrome 98.0.4
  - Other Information:
####  Stacktrace:

```
panic: send on closed channel

goroutine 369 [running]:
github.com/pion/mediadevices.(*baseTrack).unbind(0x3962393261333461, {{0xc00027c8d0, 0x20}, {{0x7ff693bfa658, 0x0, 0x0}, {0xc000f00000, 0x1, 0x1}}, 0xec9d1d4, ...})
	$GOPATH/pkg/mod/github.com/pion/mediadevices@v0.3.2/track.go:227 +0x70
github.com/pion/mediadevices.(*VideoTrack).Unbind(0x0, {{0xc00027c8d0, 0x20}, {{0x7ff693bfa658, 0x0, 0x0}, {0xc000f00000, 0x1, 0x1}}, 0xec9d1d4, ...})
	$GOPATH/pkg/mod/github.com/pion/mediadevices@v0.3.2/track.go:311 +0x45
github.com/pion/webrtc/v3.(*RTPSender).ReplaceTrack(0xc00016c2c0, {0x0, 0x0})
	$GOPATH/pkg/mod/github.com/pion/webrtc/v3@v3.1.10/rtpsender.go:157 +0x198
github.com/pion/webrtc/v3.(*RTPSender).Stop(0xc00016c2c0)
	$GOPATH/pkg/mod/github.com/pion/webrtc/v3@v3.1.10/rtpsender.go:240 +0x85
github.com/pion/webrtc/v3.(*RTPTransceiver).Stop(0xc00019a510)
	$GOPATH/pkg/mod/github.com/pion/webrtc/v3@v3.1.10/rtptransceiver.go:148 +0x7e
github.com/pion/webrtc/v3.(*PeerConnection).Close(0xc000029200)
	$GOPATH/pkg/mod/github.com/pion/webrtc/v3@v3.1.10/peerconnection.go:1878 +0x1c5
gitlab.com/awati/vrp-server/peerconnection.CloseConnection(0xc0000a3fd0, 0x23c)
	$PROJECT_FOLDER/peerconnection/peerconnection.go:360 +0x148
gitlab.com/awati/vrp-server/peerconnection.StartEngine.func1(0x5)
	$PROJECT_FOLDER/peerconnection/peerconnection.go:117 +0x48
created by github.com/pion/webrtc/v3.(*PeerConnection).onConnectionStateChange
	$GOPATH/pkg/mod/github.com/pion/webrtc/v3@v3.1.10/peerconnection.go:488 +0x175
```

### What did you do?
I call a close function when the state changes to some imcompatible status:
```go
 peerConnection.OnConnectionStateChange(func(pcs webrtc.PeerConnectionState) {
 	switch pcs {
	case webrtc.PeerConnectionStateDisconnected, webrtc.PeerConnectionStateClosed, webrtc.PeerConnectionStateFailed:
 		CloseConnection(peerConnection, pid)
 		fallthrough
 		...
```
and inside the CloseConnection i call peerConnection.Close()

### What did you expect?
To close the connection and continue to run the program
 
### What happened?
the process end with a panic message

EmrysMyrddin:
#### Description

Closing a record session on macOS does not close the reader.

This PR fixes this by closing the channel used to read incoming video frames directly in the Close function.
Calling it in the data callback is not working because since the record session is closed, the callback will never be called again. So the reader will indefinitely wait for a incoming frame/the channel to be closed, which will never happen.

By fixing this, i have also discover a bug, where closing a peer connection with some tracks already closed triggers a panic due to the stop mechanism.

#### Reference issue
Fixes #379

### DIFF
--- a/pkg/avfoundation/avfoundation_darwin.go
+++ b/pkg/avfoundation/avfoundation_darwin.go
@@ -118,8 +118,8 @@ func (rc *ReadCloser) dataCb(data []byte) {
 		return
 	}
 	select {
+	// Use the Done channel to avoid waiting for new data from closed camera
 	case <-rc.cancelCtx.Done():
-		close(rc.dataChan)
 	case rc.dataChan <- data:
 	}
 }
@@ -141,6 +141,7 @@ func (rc *ReadCloser) Close() {
 		rc.onClose()
 	}
 	rc.cancelFunc()
+	close(rc.dataChan)
 	unregister(rc.id)
 }
 
--- a/pkg/avfoundation/avfoundation_darwin.go
+++ b/pkg/avfoundation/avfoundation_darwin.go
@@ -147,9 +147,9 @@ func (rc *ReadCloser) Close() {
 		rc.onClose()
 	}
 	rc.cancelFunc()
+	unregister(rc.id)
 	rc.closeWG.Wait()
 	close(rc.dataChan)
-	unregister(rc.id)
 }
 
 // Session represents a capturing session.
--- a/pkg/avfoundation/avfoundation_darwin.go
+++ b/pkg/avfoundation/avfoundation_darwin.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sync"
 	"unsafe"
 
 	"github.com/pion/mediadevices/pkg/frame"
@@ -99,6 +100,7 @@ type ReadCloser struct {
 	onClose    func()
 	cancelCtx  context.Context
 	cancelFunc func()
+	closeWG    *sync.WaitGroup
 }
 
 func newReadCloser(onClose func()) *ReadCloser {
@@ -109,10 +111,14 @@ func newReadCloser(onClose func()) *ReadCloser {
 	cancelCtx, cancelFunc := context.WithCancel(context.Background())
 	rc.cancelCtx = cancelCtx
 	rc.cancelFunc = cancelFunc
+	rc.closeWG = &sync.WaitGroup{}
 	return &rc
 }
 
 func (rc *ReadCloser) dataCb(data []byte) {
+	rc.closeWG.Add(1)
+	defer rc.closeWG.Done()
+
 	// TODO: add a policy for slow reader
 	if rc.cancelCtx.Err() != nil {
 		return
@@ -141,6 +147,7 @@ func (rc *ReadCloser) Close() {
 		rc.onClose()
 	}
 	rc.cancelFunc()
+	rc.closeWG.Wait()
 	close(rc.dataChan)
 	unregister(rc.id)
 }
--- a/track.go
+++ b/track.go
@@ -188,7 +188,8 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
 			close(stopRead)
 			encodedReader.Close()
 
-			// When there's another call to unbind, it won't block since we mark the signalCh to be closed
+			// When there's another call to unbind, it won't block since we remove the current ctx from active connections
+			track.removeActivePeerConnection(ctx.ID())
 			close(signalCh)
 			if doneCh != nil {
 				close(doneCh)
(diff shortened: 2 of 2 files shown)


pkg/avfoundation/avfoundation_darwin.go
```diff
@@ -118,8 +118,8 @@ func (rc *ReadCloser) dataCb(data []byte) {
 		return
 	}
 	select {
+	// Use the Done channel to avoid waiting for new data from closed camera
 	case <-rc.cancelCtx.Done():
-		close(rc.dataChan)
 	case rc.dataChan <- data:
```
at-wat, line 123: Won't it cause `send to closed channel` when `Close()` while running?

EmrysMyrddin: You are right, if the context is marked as done, the result of this select is random, can be both ok or panic.
I probably just had lot of luck when tested it ^^

I will search a better way to handle this.

EmrysMyrddin: Ok, after some tests, it's not luck.

We cancel the context before closing the channel, so the `case <-rc.cancelCtx.Done():` branch will always be triggered before the write on closed channel.

I can unregister the callback before closing the channel if you want but it shouldn't change any thing since we also check for the context not being canceled at the start of the callback.

at-wat: I wrote a simplified test and it seems panic at low probability.
```go
package main

import (
  "context"
  "sync"
  "testing"
)

func TestMain(t *testing.T) {
  const n = 1000000
  var nPanic int
  for i := 0; i < n; i++ {
    func() {
      defer func() {
        if r := recover(); r != nil {
          t.Errorf("Panic: %v", r)
          nPanic++
        } 
      }()
      ch := make(chan struct{})
      ctx, cancel := context.WithCancel(context.Background())

      var wg sync.WaitGroup

      wg.Add(1)
      go func() {
        cancel()
        close(ch)
        wg.Done()
      }()

      select {
      case ch <- struct{}{}:
      case <-ctx.Done():
      }
      wg.Wait()
    }()
  }
  if nPanic > 0 {
    t.Fatalf("panic %d/%d", nPanic, n)
  } 
}
```
Around 20/1000000 on my environment

at-wat: I think this logic is used all over the pion code.
It would be ok to merge this for now to avoid larger problem, but better to have safe logic.

EmrysMyrddin: Ok, will try to avoid this.

My idea is to use a WaitingGroup to be sure every callbacks have returned and that I can safely close the channel.
What do you think ?

at-wat: Sounds good to me!

pkg/avfoundation/avfoundation_darwin.go
```diff
@@ -141,6 +147,8 @@ func (rc *ReadCloser) Close() {
 		rc.onClose()
 	}
 	rc.cancelFunc()
+	rc.closeWG.Wait()
+	close(rc.dataChan)
 	unregister(rc.id)
```
at-wat, line 152: I believe current code fixes all cases, but would be better to put unregister first to reduce unnecessary call which is aborted by `rc.cancelCtx.Err()` check.

EmrysMyrddin: Shouldn't we also close something to avoid the callback being called at all ? I mean, unregister will just silence the the fact we receive data right ?

EmrysMyrddin: Ho ok, it's because we close the avfoundation session in the code calling this close function.

at-wat (approved):
LGTM

PR #417, opened 2022-07-27
fix nil reference when a read error occur with copyFrame enabled

EmrysMyrddin:
#### Description

Fixes a nil pointer reference in the Broadcaster when using the `copyFrame` feature.
We were trying to copy the read buffer even if an error occured

### DIFF
--- a/pkg/io/video/broadcast_test.go
+++ b/pkg/io/video/broadcast_test.go
@@ -1,6 +1,7 @@
 package video
 
 import (
+	"fmt"
 	"image"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


maeln (approved):

pkg/io/video/broadcast_test.go
```diff
@@ -47,3 +48,21 @@ func TestBroadcast(t *testing.T) {
 		t.Fatal("Expected actual frame without copy to be the same with the original")
 	}
 }
+
+func TestBroadcastWithCopyOnReadError(t *testing.T) {
+	expectedError := fmt.Errorf("expected error")
```
at-wat, line 53:
```suggestion
	expectedError := errors.New("expected error")
```

EmrysMyrddin: Github Copilot seems to still have some room to improvements 😛

at-wat (approved):
LGTM

PR #418, opened 2022-07-27
fix race condition in i420 video converter leading to video artefacts

EmrysMyrddin:
#### Description

Fixes a race condition in the i420 video converter leading to artefacts when used in concurrently.

A way to see it in action is to create 2 x264 video encoder for the same video track with a video source using i422 or i444 color space. This triggers a bug where we don't convert the entire image, we only resample it, probably to avoid unecessary memory allocation. But when multiple encoders uses this converter at the same time, it will resample multiple time with the same original sample ration, leading to weird artefacts.

There is probably 2 way to fix this:
 - add locks to avoid concurrent modifications
 - make a copy when doing destructive operation

I have implemented the later one in this PR since adding a lock would be difficult, necessitating to be aware of it everywhere the frame is used. I think the simplicity to have a very local fix covers the additional memory and cpu consumption.

### DIFF
--- a/pkg/io/video/convert.go
+++ b/pkg/io/video/convert.go
@@ -15,7 +15,13 @@ func imageToYCbCr(dst *image.YCbCr, src image.Image) {
 
 	yuvImg, ok := src.(*image.YCbCr)
 	if ok {
-		*dst = *yuvImg
+		// pixel format conversion functions modify the image internals, we have to make a copy to
+		// avoid concurrent encoders to convert multiple times the same frame
+		if yuvImg.SubsampleRatio == image.YCbCrSubsampleRatio420 {
+			*dst = *yuvImg
+		} else {
+			*dst = makeCopy(*yuvImg, yuvImg.Cb, yuvImg.Cr)
+		}
 		return
 	}
 
@@ -71,16 +77,10 @@ func ToI420(r Reader) Reader {
 
 		imageToYCbCr(&yuvImg, img)
 
-		if yuvImg.SubsampleRatio == image.YCbCrSubsampleRatio420 {
-			return &yuvImg, func() {}, nil
-		}
-
-		// pixel format conversion functions modify the image internals, we have to make a copy to
-		// avoid concurrent encoders to convert multiple times the same frame
-		yuvImg := makeCopy(yuvImg, yuvImg.Cb, yuvImg.Cr)
-
 		// Covert pixel format to I420
 		switch yuvImg.SubsampleRatio {
+		case image.YCbCrSubsampleRatio420:
+			return &yuvImg, func() {}, nil
 		case image.YCbCrSubsampleRatio444:
 			i444ToI420(&yuvImg)
 		case image.YCbCrSubsampleRatio422:
--- a/pkg/io/video/convert.go
+++ b/pkg/io/video/convert.go
@@ -15,13 +15,7 @@ func imageToYCbCr(dst *image.YCbCr, src image.Image) {
 
 	yuvImg, ok := src.(*image.YCbCr)
 	if ok {
-		// pixel format conversion functions modify the image internals, we have to make a copy to
-		// avoid concurrent encoders to convert multiple times the same frame
-		if yuvImg.SubsampleRatio == image.YCbCrSubsampleRatio420 {
-			*dst = *yuvImg
-		} else {
-			*dst = makeCopy(*yuvImg, yuvImg.Cb, yuvImg.Cr)
-		}
+		*dst = *yuvImg
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/io/video/convert.go
```diff
@@ -87,6 +94,15 @@ func ToI420(r Reader) Reader {
 	})
 }
 
+func makeCopy(img image.YCbCr, cb, cr []uint8) image.YCbCr {
```
EmrysMyrddin, line 97: I don't use a pointer here to use a stack allocated copy of the YCbCr structure and related slices. This should avoid adding garbage collection pressure to the cost of the copy.

pkg/io/video/convert_test.go
```diff
@@ -222,8 +222,8 @@ func BenchmarkToI420(b *testing.B) {
 			"RGBA": image.NewRGBA(image.Rect(0, 0, sz[0], sz[1])),
 		}
 		b.Run(name, func(b *testing.B) {
-			for name, img := range cases {
-				img := img
+			for _, name := range [...]string{"I444", "I422", "I420", "RGBA"} {
```
EmrysMyrddin, line 225: map key order is not guaranteed, leading to random benchmark result order which benchstats seems to not appreciate

EmrysMyrddin:
This has a inpact on performances (+50% on 480p, +100% in 1080p). It still far better than the `FrameBuffer.StoreCopy` which converts to RGBA when we enable frame copy.

Perhaps someone has a better idea than my naive copy implementation ? perhaps something written in C ? since we already massively rely on C in this package.

at-wat:
> Perhaps someone has a better idea than my naive copy implementation ? perhaps something written in C ? since we already massively rely on C in this package.

I wrote several CGO version of heavy functions in this package, but left pure Go version and make them automatically switched by `cgo` build tag.
See
https://github.com/pion/mediadevices/blob/master/pkg/frame/yuv_nocgo.go
https://github.com/pion/mediadevices/blob/master/pkg/frame/yuv_cgo.go
for examples.

EmrysMyrddin:
After some though, I have tried an other approach, which I find cleaner: I don't copy image to avoid modification, I have refactored the converters to make them side effect free and not relying on pointers to give back the result of the conversion.

I'm much happier of this version, and the performance impact is far more contained:

```
name                  old time/op  new time/op  delta
ToI420/480p/I444-12    121µs ± 1%   169µs ± 0%  +39.17%  (p=0.008 n=5+5)
ToI420/480p/I422-12   70.3µs ± 1%  82.1µs ± 1%  +16.79%  (p=0.008 n=5+5)
ToI420/480p/I420-12   9.06ns ± 3%  9.26ns ± 1%     ~     (p=0.151 n=5+5)
ToI420/480p/RGBA-12    435µs ± 0%   480µs ± 2%  +10.46%  (p=0.008 n=5+5)
ToI420/1080p/I444-12   705µs ± 1%  1006µs ± 1%  +42.70%  (p=0.008 n=5+5)
ToI420/1080p/I422-12   405µs ± 2%   489µs ± 1%  +20.72%  (p=0.008 n=5+5)
ToI420/1080p/I420-12  9.01ns ± 0%  8.62ns ± 0%   -4.30%  (p=0.008 n=5+5)
ToI420/1080p/RGBA-12  2.72ms ± 0%  3.06ms ± 2%  +12.24%  (p=0.008 n=5+5)
```

pkg/io/video/convert_cgo.go
```diff
@@ -14,27 +15,35 @@ import "C"
 // All functions switched at runtime must be declared also in convert_nocgo.go.
 const hasCGOConvert = true
 
-func i444ToI420(img *image.YCbCr) {
+func i444ToI420(img image.YCbCr) image.YCbCr {
 	h := img.Rect.Dy()
+	cLen := img.CStride * h / 4
+	cbDst, crDst := make([]uint8, cLen), make([]uint8, cLen)
```
EmrysMyrddin, line 21: I tried to use a 0 len here with a cLen capacity to avoid the overhead of filling slice with empty values.

But I didn't found any way to get the data stored by the C code. Keep having index out of range, and didn't found any way to set the slice len without losing the data.

at-wat: `sync.Pool` should reduce such overhead.

EmrysMyrddin: I will try to see if it helps to speed up things

EmrysMyrddin: Ok, I have tried to use Pool, but since the goal is to avoid the share of the slice between runs, I didn't find any way to use it to improve performance. It always involve at some point making a slice of the right length.

EmrysMyrddin: Perhpas we can merge this as is and we will see in the future if we find any good solution ?
When the release function will be actually used for example ?

at-wat: Release function in example:
https://github.com/pion/mediadevices/blob/43272ea9658e2b75b7e382e9f71d20f1e39ece2b/examples/http/main.go#L62-L71

I think it's ok to merge and optimize later.

at-wat (approved):

PR #419, opened 2022-07-29
allow to close the avfoundation device multiple time

EmrysMyrddin:
#### Description

Allow to close multiple time an avfoundation. For now, when closing an alreacy closed VideoTrack comming from avfoundation leads to a panic (close of closed channel)

### DIFF
--- a/pkg/avfoundation/avfoundation_darwin.go
+++ b/pkg/avfoundation/avfoundation_darwin.go
@@ -101,6 +101,7 @@ type ReadCloser struct {
 	cancelCtx  context.Context
 	cancelFunc func()
 	closeWG    *sync.WaitGroup
+	lock       *sync.Mutex
 }
 
 func newReadCloser(onClose func()) *ReadCloser {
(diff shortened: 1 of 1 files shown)


pkg/avfoundation/avfoundation_darwin.go
```diff
@@ -101,6 +101,7 @@ type ReadCloser struct {
 	cancelCtx  context.Context
 	cancelFunc func()
 	closeWG    *sync.WaitGroup
+	lock       *sync.Mutex
```
at-wat, line 104:
```suggestion
	lock       sync.Mutex
```
then we don't need to initialize like `rc.lock = &sync.Mutex{}`

(above WaitGroup as well)

pkg/avfoundation/avfoundation_darwin.go
```diff
@@ -156,11 +163,13 @@ func (rc *ReadCloser) Close() {
...
 	device   Device
 	cSession C.PAVBindSession
+	lock     sync.Mutex
+	closed   bool
 }
 
 // NewSession creates a new capturing session
 func NewSession(device Device) (*Session, error) {
-	var session Session
+	session := Session{}
```
at-wat, line 172:
```suggestion
	var session Session
```

at-wat (approved):

PR #420, opened 2022-07-29
fix PLI and FIR handling, wrongly triggering track.OnEnded

EmrysMyrddin:
#### Description

Fix the newly added PLI and FIR handling. An error has been introduce when merging the PR.

We were reading with an empty buffer, which lead to `io.ErrShortBuffer` error. This was causing the RTCP read loop to crash, but the track remained sending video, since the reading and writing loops were not linked.

A potentially breaking side effect was also the call to `track.OnEnded`, while the track didn't realy ended in facts.

I'm fixing this by using a buffer of 1500 bytes (from what I have found in different examples), and closing the track when read errors occurs.

### DIFF
--- a/track.go
+++ b/track.go
@@ -231,6 +231,8 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
 
 func (track *baseTrack) rtcpReadLoop(reader interceptor.RTCPReader, keyFrameController codec.KeyFrameController, stopRead chan struct{}) {
 	readerBuffer := make([]byte, rtcpInboundMTU)
+
+readLoop:
 	for {
 		select {
 		case <-stopRead:
@@ -240,13 +242,17 @@ func (track *baseTrack) rtcpReadLoop(reader interceptor.RTCPReader, keyFrameCont
 
 		readLength, _, err := reader.Read(readerBuffer, interceptor.Attributes{})
 		if err != nil {
-			track.onError(err)
-			return
+			if errors.Is(err, io.EOF) {
+				return
+			}
+			logger.Warnf("failed to read rtcp packet: %s", err)
+			continue
 		}
 
 		pkts, err := rtcp.Unmarshal(readerBuffer[:readLength])
 		if err != nil {
 			logger.Warnf("failed to unmarshal rtcp packet: %s", err)
+			continue
 		}
 
 		for _, pkt := range pkts {
@@ -254,6 +260,7 @@ func (track *baseTrack) rtcpReadLoop(reader interceptor.RTCPReader, keyFrameCont
 			case *rtcp.PictureLossIndication, *rtcp.FullIntraRequest:
 				if err := keyFrameController.ForceKeyFrame(); err != nil {
 					logger.Warnf("failed to force key frame: %s", err)
+					continue readLoop
 				}
 			}
 		}
--- a/track.go
+++ b/track.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	rtpOutboundMTU = 1200
+	rtcpInboundMTU = 1500
 )
 
 var (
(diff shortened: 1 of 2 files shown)


at-wat:
Is it possible to add a unit test?

EmrysMyrddin:
Sure, do you have any suggestion on how to implement this tests ? Making an RTCPReader mock ?

track.go
```diff
@@ -222,38 +223,43 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
...
-				var readerBuffer []byte
-				_, _, err := reader.Read(readerBuffer, interceptor.Attributes{})
-				if err != nil {
-					track.onError(err)
-					return
-				}
+	return selectedCodec, nil
+}
+
+func (track *baseTrack) rtcpReadLoop(reader interceptor.RTCPReader, keyFrameController codec.KeyFrameController, stopRead chan struct{}) {
```
EmrysMyrddin, line 232: Extracting the RTCP handler loop allow me to more easily unit test this part of the code.

The easiest way is to mock the RTCPReader since it is a mono function interface.

track_test.go
```diff
@@ -53,3 +55,101 @@ func TestOnEnded(t *testing.T) {
...
+}
+
+type fakeKeyFrameController struct {
+	called chan struct{}
+}
+
+func (mock *fakeKeyFrameController) ForceKeyFrame() error {
+	mock.called <- struct{}{}
+	return nil
+}
```
EmrysMyrddin, line 84: Not sure where to put this ? Do you want to push this to the end of the file since they are just test utils ?

EmrysMyrddin:
I have implemented a unit test to check for good PLI and FIR detection. I have mocked the RTCPReader, which I think was the more easy and robust way to test this code. But the tradeoff is that the read buffer size bug would not have been detected with this test. Or at least not exactly the same bugs (the test requires a buffer size of 28 bytes minimum).

I haven't found the official maximum packet size used by pion (I'm using the size used in tests and examples). So I don't want to add a test for this.

track.go
```diff
@@ -223,41 +223,43 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
...
+		readLength, _, err := reader.Read(readerBuffer, interceptor.Attributes{})
+		if err != nil {
+			track.onError(err)
+			return
+		}
+
+		pkts, err := rtcp.Unmarshal(readerBuffer[:readLength])
+		if err != nil {
+			track.onError(err)
+			return
```
at-wat, line 250: I think it's better to continue the read loop on RTCP packet processing error.
As RTCP is on UDP, sender won't know receiver stopped reading and will continue sending packets.

EmrysMyrddin: I agree. I should also probably not exit the read loop if the ForceKeyFrame failed.

Perhaps adding a log to avoid silently discarding a potential problem ?

track.go
```diff
@@ -222,38 +223,41 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
...
+		case <-stopRead:
+			return
+		default:
+		}
 
-				pkts, err := rtcp.Unmarshal(readerBuffer)
+		readLength, _, err := reader.Read(readerBuffer, interceptor.Attributes{})
+		if err != nil {
+			track.onError(err)
+			return
```
at-wat, line 244: Maybe this is also recoverable?

at-wat:
> I haven't found the official maximum packet size used by pion (I'm using the size used in tests and examples).

Maximum packet size is limited by Ethernet frame size (1500 bytes).
Possible exception is Ehternet jumbo frame, but at least browsers don't use it.
I'm not very sure there are any webrtc implementation using jumbo frame.

I think it's fine to use 1500 and consider jumbo frame later when requested.

at-wat (approved):
LGTM

PR #368, opened 2021-11-23
Add darwin arm64 libs

at-wat:

PR #428, opened 2022-08-06
Update libopus static library to correct arch for darwin arm64

edaniels:
See https://github.com/pion/mediadevices/pull/368#issuecomment-1114623571


Sean-Der:
@edaniels This is just removing, did you also intend to add the correct one?

thanks for opening this PR!

edaniels:
@Sean-Der It's a modification! Curious that it's not an add and remove since it's binary.

at-wat:
@edaniels Could you update
https://github.com/pion/mediadevices/blob/master/dockerfiles/darwin-arm64.cmake
https://github.com/pion/mediadevices/blob/master/dockerfiles/darwin-arm64.Dockerfile
to make it cross-compilable?
All other static libraries are built by the scripts in https://github.com/pion/mediadevices/tree/master/dockerfiles and it's very hard to be maintained in the future without cross-compilation.

edaniels:
I did this via the make target so there's no changes to make. It looks like the wrong file was originally committed :)

at-wat:
Is `pkg/codec/openh264/lib/libopenh264-darwin-arm64.a` incorrect as well?

edaniels:
lookin good 8)
```
Non-fat file: ./pkg/codec/openh264/lib/libopenh264-darwin-arm64.a is architecture: arm64
Non-fat file: ./pkg/codec/openh264/lib/libopenh264-darwin-x64.a is architecture: x86_64
```

at-wat (approved):
LGTM:+1:

PR #436, opened 2022-08-08
Properly close microphone before malgo chunk channel

edaniels:
Prior to this, there's a chance you would close the chunk channel while malgo is trying to send on it. Instead, it's better to stop the malgo device and then close. That also means we need to discard chunks when trying to stop due to the buffered channel.

### DIFF
--- a/pkg/driver/microphone/microphone.go
+++ b/pkg/driver/microphone/microphone.go
@@ -1,6 +1,7 @@
 package microphone
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -126,43 +127,33 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
 		return nil, errUnsupportedFormat
 	}
 
+	cancelCtx, cancel := context.WithCancel(context.Background())
 	onRecvChunk := func(_, chunk []byte, framecount uint32) {
-		m.chunkChan <- chunk
+		select {
+		case <-cancelCtx.Done():
+		case m.chunkChan <- chunk:
+		}
 	}
 	callbacks.Data = onRecvChunk
 
 	device, err := malgo.InitDevice(ctx.Context, config, callbacks)
 	if err != nil {
+		cancel()
 		return nil, err
 	}
 
 	err = device.Start()
 	if err != nil {
+		cancel()
 		return nil, err
 	}
 
 	var closeDeviceOnce sync.Once
 	m.deviceCloseFunc = func() {
 		closeDeviceOnce.Do(func() {
-			closeDone := make(chan struct{})
-			// we may be waiting on a chunk so keep reading until we can stop
-			// the malgo device.
-			go func() {
-				for {
-					select {
-					case <-closeDone:
-						return
-					default:
-					}
-					select {
-					case <-m.chunkChan:
-					case <-closeDone:
-					}
-				}
-			}()
+			cancel() // Unblock onRecvChunk
 			device.Stop()
 			device.Uninit()
-			close(closeDone)
 		})
 	}
 
--- a/pkg/driver/microphone/microphone.go
+++ b/pkg/driver/microphone/microphone.go
@@ -93,10 +93,6 @@ func (m *microphone) Close() error {
 	if m.deviceCloseFunc != nil {
 		m.deviceCloseFunc()
 	}
-	if m.chunkChan != nil {
-		close(m.chunkChan)
-		m.chunkChan = nil
-	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/driver/microphone/microphone.go
```diff
@@ -136,11 +141,35 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
...
+				for {
+					select {
+					case <-closeDone:
+						return
+					default:
+					}
+					select {
+					case <-m.chunkChan:
+					case <-closeDone:
+					}
```
at-wat, line 160: Is `<-m.chunkChan` always received after `<-closeDone`?
I guess `m.chunkChan` and `closeDone` may be triggered almost at the same time and may return before consuming `m.chunkChan` at low chance.

```go
func (m *microphone) Close() error {
	if m.deviceCloseFunc != nil {
		m.deviceCloseFunc()
	}
	// Don't close m.chunkChan to avoid panic due to `send to closed chan`
	// Go GC collects unclosed chan with buffered value when unused:
	//   https://groups.google.com/g/golang-nuts/c/KtIyc5lTRJY/m/6PK6aK5QKHIJ
	return nil
}
...
func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
	...
	ctx, cancel := context.WithCancel(context.Background())
	onRecvChunk := func(_, chunk []byte, framecount uint32) {
		select {
		case m.chunkChan <- chunk:
		case <-ctx.Done():
		}
	}
	...
	m.deviceCloseFunc = func() {
		closeDeviceOnce.Do(func() {
			cancel() // Unblock onRecvChunk
			device.Stop()
			device.Uninit()
		})
	}
```
would be more safe?

edaniels: Yup, good catch. Will use your version. The whole point is to unblock anyway, not consume.

edaniels: Also check your gophers slack!

edaniels: I'll still use your way but the original way is also safe because closeDone only happens after we stop the device which means the callback will never fire and no one will get stuck.

edaniels:
@at-wat updated. I left the intentional close in to not rely on an GC implementation detail

pkg/driver/microphone/microphone.go
```diff
@@ -87,6 +90,9 @@ func (m *microphone) Open() error {
 }
 
 func (m *microphone) Close() error {
+	if m.deviceCloseFunc != nil {
+		m.deviceCloseFunc()
+	}
 	if m.chunkChan != nil {
 		close(m.chunkChan)
```
at-wat, line 97: I think this can be called at exactly same timing with `m.chunkChan <- chunk` and cause panic by low chance.
(same as https://github.com/pion/mediadevices/pull/414#discussion_r930515146)

edaniels: I don't think it can be. The close function calls Stop/Uninit (turns out only Uninit is required). Both of those functions wait for the worker thread to return and ultimately will no longer call the stop/data callbacks. Therefore there are no more sends on the channel after and we can safely close it. Additionally you can see the Remarks section in https://miniaudio.docsforge.com/master/api/#ma_device_callback_proc referring to deadlocks and https://miniaudio.docsforge.com/master/api/ma_device_uninit/ showing the worker thread being waited on to terminate.

at-wat: Oh I see, as long as the uninit waits for ongoing callback return, it should be safe.

Maybe moving `close(m.chunkChan)` into `deviceCloseFunc` would be better at the readability about this safety logic.

edaniels: moved!

pkg/driver/microphone/microphone.go
```diff
@@ -153,6 +149,11 @@ func (m *microphone) AudioRecord(inputProp prop.Media) (audio.Reader, error) {
 		closeDeviceOnce.Do(func() {
 			cancel() // Unblock onRecvChunk
 			device.Uninit()
+
+			if m.chunkChan != nil {
+				close(m.chunkChan)
+				m.chunkChan = nil
+			}
```
at-wat, line 156: As it's in Once,
```suggestion
			close(m.chunkChan)
			m.chunkChan = nil
```

at-wat (approved):
LGTM!

pkg/driver/microphone/microphone.go
edaniels: I left that there because we don't check if we've called Open, so the channel may be nil, hypothetically

PR #437, opened 2022-08-16
force key frame implementation for vaapi

EmrysMyrddin:
#### Description

Force key frame implementation for the VAAPI in both VP8 and VP9 mode.


pkg/codec/vaapi/vp8.go
```diff
@@ -545,6 +547,10 @@ func (e *encoderVP8) Controller() codec.EncoderController {
 	return e
 }
 
+func (e *encoderVP8) ForceKeyFrame() {
```
at-wat, line 550: This interface has been replaced by `func Controller() codec.EncoderController`
https://github.com/pion/mediadevices/pull/338/files#diff-f1af839aa40e8d1553814897ef277f31f625ed6daf6e6f44bfeb950905efabb8R122-R146

EmrysMyrddin: I don't understand, this function method already exists and returns the encoder itself.

at-wat: I also don't remember what I wanted to point and old CI log is gone.
The interface seems be correct.

pkg/codec/vaapi/vp8.go
```diff
@@ -545,6 +547,10 @@ func (e *encoderVP8) Controller() codec.EncoderController {
...
+	e.forceKeyFrame = true
```
at-wat, line 551: It may cause data race if `ForceKeyFrame()` is called from other goroutine during `Read()`?

pkg/codec/vaapi/vp8.go
```diff
@@ -315,7 +317,7 @@ func (e *encoderVP8) Read() ([]byte, func(), error) {
 
 	e.frParam.data.framerate = C.uint(e.rate.Calc())
 
-	if kf {
+	if kf || e.forceKeyFrame {
```
at-wat, line 320: `forceKeyFrame` isn't reset, so every frames after `ForceKeyFrame()` will be keyframe.

pkg/codec/vaapi/vp8.go
EmrysMyrddin: I've changed the type to use an `atomic.Bool` instead. This way, there is no data race possible.

EmrysMyrddin:
@edaniels I've merged master and applied suggestions if you want :-)

edaniels (approved):
LGTM. Thanks!

pkg/codec/vaapi/vp8.go
```diff
@@ -100,6 +101,8 @@ type encoderVP8 struct {
 
 	rate *framerateDetector
 
+	forceKeyFrame atomic.Bool
```
at-wat, line 104: Need to bump minimal supported go version to 1.19
https://github.com/pion/mediadevices/blob/b4770e5fbf35214a92cb0a5fc335283893bd9fa8/go.mod#L3

PR #453, opened 2022-11-12
Add missing EnablePictureID in VP8Payloader

vpalmisano:
The EnablePictureID param is required to create the extended RTP VP8 header. Without this header the VP8 streams cannot be read by the Mediasoup SFU.


at-wat (commented):
It's better to be configurable, but seems difficult on the current API.
I think having a global switch like
```go
var VP8EnablePictureID = true
```
would be good for now.
Enabled by default should be fine.

PR #455, opened 2022-11-21
upgrade malgo to v0.11.10

edaniels:
brings a host of fixes and gets rid of a pesky issue on macos


edaniels:
Tested against https://github.com/edaniels/gostream/blob/master/cmd/stream_audio/main.go and it works great still

PR #463, opened 2023-01-09
Expose device list update func

bazile-clyde:
#### Description
This PR exports the logic executed at startup to find and register cameras. This will allow us to discover cameras connected after our application has started.

### DIFF
--- a/pkg/driver/camera/camera_darwin.go
+++ b/pkg/driver/camera/camera_darwin.go
@@ -17,6 +17,10 @@ type camera struct {
 }
 
 func init() {
+	Initialize()
+}
+
+func Initialize() {
 	devices, err := avfoundation.Devices(avfoundation.Video)
 	if err != nil {
 		panic(err)
--- a/pkg/driver/camera/camera_darwin.go
+++ b/pkg/driver/camera/camera_darwin.go
@@ -20,6 +20,7 @@ func init() {
 	Initialize()
 }
 
+// Initialize finds and registers camera devices. This is part of an experimental API.
 func Initialize() {
 	devices, err := avfoundation.Devices(avfoundation.Video)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


at-wat:
Do you think we can expose setup function in other drivers as well?
It would be better to have unified interface to update device list.

bazile-clyde:
> Do you think we can expose setup function in other drivers as well?
> It would be better to have unified interface to update device list.

@at-wat Done! I skipped the `dummy.go` files in `audiotest` and `videotest`. I also skipped `vncdriver.go`  since there's no `func init`. Let me know if I should _not_ have skipped any of those.

at-wat:
@bazile-clyde thanks!

I think it would be better in future to have another API that `mediadevices.MediaDevices` can automatically call them to update the device list.
For this, `Initialize` functions would be better to have godoc comments to mention this API is experimental.
What do you think?

bazile-clyde:
> @bazile-clyde thanks!
> 
> I think it would be better in future to have another API that `mediadevices.MediaDevices` can automatically call them to update the device list. For this, `Initialize` functions would be better to have godoc comments to mention this API is experimental. What do you think?

That makes sense. I don't have strong opinions either way so I added it. Thanks for the quick turnaround btw!

at-wat (changes requested):
Could you revert import block orders to keep them like
```go
import (
  "stdlib1"
  "stdlib2"

  "nonstdlib1"
  "nonstdlib2"
)
```

bazile-clyde:
> Could you revert import block orders to keep them like
> 
> ```go
> import (
>   "stdlib1"
>   "stdlib2"
> 
>   "nonstdlib1"
>   "nonstdlib2"
> )
> ```

Done!

at-wat (approved):
LGTM!

Issue #407, opened 2022-06-14
Upgrade for Pi OS


DevelopDaily:
## Summary

Upgrade for Pi OS

## Motivation

Has anybody planned to upgrade the project for the latest Pi OS? Raspberry Pi has upgraded from Raspbian to Pi OS. They said they would discontinue the "Legacy Camera", which `pion/mediadevices` relies on right now, in the future.

This project is so useful to the Raspberry Pi ecosystem that a thorough upgrade for Pi OS would benefit a lot of new apps.

Right now, the `webrtc `example does not work on Pi OS. I have to rebuild the opus library to make it work.

## Describe alternatives you've considered

As I stated above, the "Legacy Camera" must be enabled and some libraries such as `opus `must be rebuilt.

![pi](https://user-images.githubusercontent.com/54299975/173480611-bc89506b-1153-4edc-b783-3cca3ba44081.jpg)

bprfh:
The problem is that libcamera doesn't present a v4l2 device, libraries are expected to call the c++ api.

To support older v4l2 based programms you can use libcamerify <bin> which uses  LD_PRELOAD to hook the open and v4l2 libc syscalls.

This doesn't work if you use the go builtin syscalls as these use assembly to directly call in the kernel instead of using libc.

The solution is to use cgo and use a bridge function for open/close/v4l2 syscalls and run the progamm.

I have a working solution, but I'm not ready to upload it yet.

You could also replace the ioctls with cgo code in the blackjack/webcam package.

PR #465, opened 2023-01-15
Add shell command media source drivers + misc improvements

KW-M:
#### Description

Added feature:
- New drivers to allow the use of shell command output as a video or audio source. Added tests and example code for this driver. This can be used in cases where command line programs support media sources that pion/mediadevices doesn't yet (EG: libcamera based cameras), or to add filters to a video or audio stream before pion/mediadevices using ffmpeg or gstreamer.

Improvements:
- Added bit-rate control to vpx and h264 encoders.
- Bit-rate is auto-scaled based on the REMB bandwidth messages now.

Fixes:
- Possible deadlock when a driver throws a read error on the very first read (see the comment on the commit [here](https://github.com/kw-m/mediadevices/commit/39923b2ca9686f42b3692556ae8d6f5e3516901d) for an example of this deadlock happening)

#### Reference issue
Workaround for #407


at-wat:
@KW-M thank you for your contribution!
Could you split PRs per feature/improvement/fix?

KW-M:
Thanks! Will do.

PR #466, opened 2023-01-20
Fix deadlock in track.Bind()

KW-M:
#### Description

Occurs when read errors happen from a
driver source during a call to track.Bind()

Example of deadlock:
1. When a driver reader function has a read error immediately after being initialized: contrived example:
```golang
video.ReaderFunc(func() (image.Image, func(), error) {
	return nil, nil, errors.New("Ahhhh")
})
```

2. Pion calls bind() on a track using that reader function.
3. Within NewRTPReader():
    a. The reader function is wrapped to call track.onError if a read error is returned in newVideoTrackFromReader()
    b. Read is called synchronously in detectCurrentVideoProp() in the function newEncodedReader()
    c.  track.onError() is triggered within the bind function (so without this fix, the mutex lock on bind() will not yet be released, and would deadlock in track.onError())

### DIFF
--- a/track.go
+++ b/track.go
@@ -172,10 +172,14 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
 	for _, wantedCodec := range ctx.CodecParameters() {
 		logger.Debugf("trying to build %s rtp reader", wantedCodec.MimeType)
 		encodedReader, err = specializedTrack.NewRTPReader(wantedCodec.MimeType, uint32(ctx.SSRC()), rtpOutboundMTU)
+
+		track.errMu.Lock()
 		if track.err != nil {
 			err = track.err
 			encodedReader = nil
 		}
+		track.errMu.Unlock()
+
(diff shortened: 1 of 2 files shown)


track.go
```diff
@@ -171,6 +172,10 @@ func (track *baseTrack) bind(ctx webrtc.TrackLocalContext, specializedTrack Trac
 	for _, wantedCodec := range ctx.CodecParameters() {
 		logger.Debugf("trying to build %s rtp reader", wantedCodec.MimeType)
 		encodedReader, err = specializedTrack.NewRTPReader(wantedCodec.MimeType, uint32(ctx.SSRC()), rtpOutboundMTU)
+		if track.err != nil {
+			err = track.err
```
at-wat, line 176: I think this block also need `track.errMu.Lock()`

KW-M: Thanks for noticing! I added the lock so It should be good now.

at-wat (approved):
LGTM

PR #467, opened 2023-01-20
Implement bitrate controller in vpx and h264 codecs

KW-M:
#### Description

- Implemented the BitRateController interface for the vpx and h264 codecs
- Added very simple REMB based bit rate handling.
This sets every track to use roughly all (hardcoded to 93%) the estimated available bandwidth.
Currently it doesn't compensate for when other tracks or data channels are transmitting over the same bandwidth. 
I couldn't find an easy way to detect what other tracks & datachannels were in use over the same peer connection, so doing bit rate control in a smarter way is still an open problem (see rtcpReadLoop in track.go).


EmrysMyrddin:
Hi, thank you for your PR, it looks awesome !

While automatic bitrate adjustment is a great feature, it should probably be configurable. We should be able to at least disable it by default and chose the percentage of the detected maximum bandwidth.

Perhaps you could split this PR, letting us merge your bitrate controllers and discuss about how to implement the automatic bitrate adjustment in a dedicated PR ?

KW-M:
Thanks for the suggestions @EmrysMyrddin. I rolled back this PR to just include the bit-rate encoder controllers.

I'm thinking the place to implement the option for auto bit-rate control are in the RTPCodec struct at the beginning of codec.go, and then maybe allow setting the percentage bandwidth used per stream by passing the percentage as a mediaStreamConstraint in getUserMedia(). Does that seem good, or might there be a better way to do it?

pkg/codec/x264/bridge.h
```diff
@@ -78,6 +79,22 @@ Encoder *enc_new(x264_param_t param, char *preset, int *rc) {
   return NULL;
 }
 
+#define RC_MARGIN 10000 /* 1kilobits / second*/
```
EmrysMyrddin, line 82: Isn't this 10 kb/s ?

EmrysMyrddin (commented):
LGTM, just a little question and we should be able to merge this :-)

3DRX (approved):
LGTM, I think we can merge this for now. I'll open future PRs to address any issues if needed, for now, I've tested this patch with no regressions found.

PR #470, opened 2023-02-02
discard buffered frames for < 1fps

bazile-clyde:
#### Problem Description
This library sets a buffer size of 1 upon opening cameras in Linux. This will always result in stale frames. For streaming this isn't a big deal since the staleness of a frame could be as little as about 30ms assuming a camera is capable of streaming at 30fps or even less for higher frame rates. 

For taking snapshots this is much more problematic. Consider the following use case: a user wants to take a snapshot once a day at noon. Every image would be a day behind. That is, Monday's snapshot would be from Sunday, Sunday's snapshot would be from Saturday, and so forth.

Simply making an extra call in the application wrapping this one is difficult since 
1. the size of the buffer differs by OS 
2. the wrapping application would then depend on the buffer size, which may change over time
3. the reader function this library returns does not distinguish between snapshots and streams.

I do not think it’s possible to add unbuffered support to blackjack/webcam either. According to the V4L spec ([link](https://www.linuxtv.org/downloads/legacy/video4linux/API/V4L2_API/spec-single/v4l2.html#io)) there are three ways to read or stream frames from cameras:
1. [Read/Write](https://www.linuxtv.org/downloads/legacy/video4linux/API/V4L2_API/spec-single/v4l2.html#rw)
2. [Streaming I/O (Memory Mapping)](https://www.linuxtv.org/downloads/legacy/video4linux/API/V4L2_API/spec-single/v4l2.html#mmap)
3. [Streaming I/O (User Pointers)](https://www.linuxtv.org/downloads/legacy/video4linux/API/V4L2_API/spec-single/v4l2.html#userp)

Method 2 is what blackjack uses currently. Both methods 2 and 3 require the allocation of buffers. Method 1 doesn’t use buffers but it isn’t as widely supported as 2 and usually not supported for USB webcams. To further drive home this point, the following is what the V4L spec says about reading frames from cameras with read/write functions

>It is considered inferior though because no meta-information like frame counters or timestamps are passed. This information is necessary to recognize frame dropping and to synchronize with other data streams. 

In short, read/write is bad and isn’t widely supported. Streaming using mmap is good and is widely supported.

#### Solution
To fix this issue I made a simple assumption: if the user is reading frames at <1fps they are taking snapshots and not streaming. The solution then is to save a timestamp of the last time we read a frame from the camera. If it was more than a second ago, clear the buffer. If it was less than a second ago, it's business as usual. This adds virtually no overhead for streaming but does add a small amount for taking snapshots. Testing this locally with a C270 USB webcam resulted in <4ms of additional latency per snapshot. If you're taking snapshots every second or greater this extra overhead is negligible.

If there are better ways to fix this problem please let me know. I also plan to investigate ways to fix the same issue on Darwin which has a buffer size of 3.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/blackjack/webcam"
 	"github.com/pion/mediadevices/pkg/driver"
@@ -23,6 +24,7 @@ import (
 const (
 	maxEmptyFrameCount = 5
 	prioritizedDevice  = "video0"
+	bufferedFrameCount = 1
 )
 
 var (
@@ -70,6 +72,7 @@ type camera struct {
 	started         bool
 	mutex           sync.Mutex
 	cancel          func()
+	prevFrameTime   time.Time
 }
 
 func init() {
@@ -158,8 +161,13 @@ func (c *camera) Open() error {
 		return err
 	}
 
-	// Late frames should be discarded. Buffering should be handled in higher level.
-	cam.SetBufferCount(1)
+	// Buffering should be handled in higher level.
+	err = cam.SetBufferCount(bufferedFrameCount)
+	if err != nil {
+		return err
+	}
+
+	c.prevFrameTime = time.Now()
 	c.cam = cam
 	return nil
 }
@@ -216,7 +224,7 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	c.cancel = cancel
 	var buf []byte
-	r := video.ReaderFunc(func() (img image.Image, release func(), err error) {
+	readerFunc := func() (img image.Image, release func(), err error) {
 		// Lock to avoid accessing the buffer after StopStreaming()
 		c.mutex.Lock()
 		defer c.mutex.Unlock()
@@ -262,6 +270,20 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 			return decoder.Decode(buf[:n], p.Width, p.Height)
 		}
 		return nil, func() {}, errEmptyFrame
+	}
+
+	r := video.ReaderFunc(func() (img image.Image, release func(), err error) {
+		// If frames are requested at less than 1fps, assume snapshots, not streaming, and discard all
+		// buffered frames to reduce staleness.
+		if time.Now().Sub(c.prevFrameTime).Seconds() >= 1 {
+			for toDiscard := bufferedFrameCount; toDiscard > 0; toDiscard-- {
+				if _, _, err = readerFunc(); err != nil {
+					return nil, func() {}, err
+				}
+			}
+		}
+		c.prevFrameTime = time.Now()
+		return readerFunc()
 	})
 
 	return r, nil


pkg/driver/camera/camera_linux.go
```diff
@@ -158,8 +161,13 @@ func (c *camera) Open() error {
 		return err
 	}
 
-	// Late frames should be discarded. Buffering should be handled in higher level.
-	cam.SetBufferCount(1)
+	// Buffering should be handled in higher level.
```
bazile-clyde, line 164: Removed "Late frames should be discarded." since we this PR discards late frames.

pkg/driver/camera/camera_linux.go
```diff
@@ -158,8 +161,13 @@ func (c *camera) Open() error {
...
+	err = cam.SetBufferCount(bufferedFrameCount)
+	if err != nil {
+		return err
+	}
+
+	c.prevFrameTime = time.Now()
```
bazile-clyde, line 170: This will cause the first call to the readerFunction to discard a stale frame if needed. That is, a frame from more than a second ago. For streaming at >1fps this will only happen once to get the stream on track then never again.

pkg/driver/camera/camera_linux.go
```diff
@@ -23,6 +24,7 @@ import (
 const (
 	maxEmptyFrameCount = 5
 	prioritizedDevice  = "video0"
+	bufferedFrameCount = 1
```
edaniels, line 27: instead of making this a const, we know that we don't really ever want to be buffering more than this, so I think we can keep the 1 below

pkg/driver/camera/camera_linux.go
edaniels, line 170: Ideally we only call this when the user wants to use this behavior

pkg/driver/camera/camera_linux.go
```diff
@@ -262,6 +270,20 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 			return decoder.Decode(buf[:n], p.Width, p.Height)
 		}
 		return nil, func() {}, errEmptyFrame
+	}
+
+	r := video.ReaderFunc(func() (img image.Image, release func(), err error) {
+		// If frames are requested at less than 1fps, assume snapshots, not streaming, and discard all
```
edaniels, line 276: I think you can factor this logic into the above

pkg/driver/camera/camera_linux.go
```diff
@@ -70,6 +72,7 @@ type camera struct {
 	started         bool
 	mutex           sync.Mutex
 	cancel          func()
+	prevFrameTime   time.Time
```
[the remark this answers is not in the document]
bazile-clyde: I like that idea! I'm not sure how it'd be best implemented but I have a few ideas:

1. We could make this struct public so users can cast their `Driver` interface into this concrete type and set `DiscardFramesOlderThan`.
2. We could add a generic method for setting properties to the `Driver` interface. Something like `Set(prop map[string]interface{})`
3. Add a non-const global variable.

I think 2 would make the most sense `Driver` isn't camera specific. All would require locking the mutex which isn't exposed, so 1 and 3 seems dangerous unless you expose a mutex.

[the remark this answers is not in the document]
edaniels: prop.Video would work as the input to this function.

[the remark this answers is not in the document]
edaniels: I don't think it needs to be a constraint since that's a selection criteria.

pkg/driver/camera/camera_linux.go
edaniels: Fair enough!

pkg/driver/camera/camera_linux.go
```diff
@@ -243,6 +260,7 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 				// Camera has been stopped.
 				return nil, func() {}, err
 			}
+			c.prevFrameTime = time.Now()
```
edaniels, line 263: let's do this if `			if p.DiscardFramesOlderThan != 0 {`

edaniels (changes requested):
Think we can remove that loop

pkg/driver/camera/camera_linux.go
bazile-clyde: There's a comment above `SetBufferCount` that suggest it may be eliminated in the future so it may not always be 1 and might go away. If that happens then the context will be lost for why we're flushing a single frame. Either way, I don't have a strong opinion on this so removed.

pkg/driver/camera/camera_linux.go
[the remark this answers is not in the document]
bazile-clyde: `prop.Video` and `prop.VideoConstraints` are merged in `prop.go::merge` using reflections. This is only possible because the two structs are identical. If they differ we'll get a `reflect: Field index out of range` error unless we handle this field specially in that function.

pkg/driver/camera/camera_linux.go
bazile-clyde: Done.

pkg/driver/camera/camera_linux.go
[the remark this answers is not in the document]
edaniels: oh lame, okay!

edaniels (approved):

PR #482, opened 2023-03-17
change buffer size to 2

bazile-clyde:
#### Description
From the Linux kernel documentation [here](https://linuxtv.org/downloads/v4l-dvb-apis/driver-api/v4l2-videobuf.html)
> As a practical rule, a minimum of two buffers are needed for proper streaming, and there is usually a maximum (which cannot exceed 32) which makes sense for each device.

This code updates the buffer size from 1 to 2 along with the frame skip logic. The side effect is much higher fps on all 3 of the cameras I've tested, presumably because the buffer is no longer a bottleneck, and there shouldn't be a risk of slowing down any streams.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -61,6 +61,8 @@ var (
 	}
 )
 
+const bufCount = 2
+
 // Camera implementation using v4l2
 // Reference: https://linuxtv.org/downloads/v4l-dvb-apis/uapi/v4l/videodev.html#videodev
 type camera struct {
@@ -162,7 +164,7 @@ func (c *camera) Open() error {
 	}
 
 	// Buffering should be handled in higher level.
-	err = cam.SetBufferCount(1)
+	err = cam.SetBufferCount(bufCount)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/driver/camera/camera_linux.go
```diff
@@ -72,6 +72,7 @@ type camera struct {
 	mutex           sync.Mutex
 	cancel          func()
 	prevFrameTime   time.Time
+	bufCount        int
```
at-wat, line 75: Is there a reason it's not
```go
const bufCount = 2
```
?

bazile-clyde: No, that's better. Done!

pkg/driver/camera/camera_linux.go
```diff
@@ -162,7 +164,7 @@ func (c *camera) Open() error {
 	}
 
 	// Buffering should be handled in higher level.
-	err = cam.SetBufferCount(1)
+	err = cam.SetBufferCount(uint32(bufCount))
```
at-wat, line 167: Const integer in Go is not typed by default, so
```suggestion
	err = cam.SetBufferCount(bufCount)
```

bazile-clyde: Ah, gotcha. Done!

at-wat (approved):
LGTM:+1:

PR #485, opened 2023-03-24
handle race conditions + add test for Manager.Register

bazile-clyde:
#### Description
I've run into race conditions accessing `Manager` concurrently. This PR adds tests and fixes for race conditions.


at-wat (approved):
LGTM:+1:

PR #486, opened 2023-03-31
fix cLen in nocgo implementation of i444Toi420

neversi:
#### Description
Fix the length of buffer - cLen in `i444toI420()` in nocgo implementation


neversi:
While optimizing copy of buffers in nocgo implementation, I have faced with this bug.

neversi:
@edaniels could you please at this bug?

edaniels:
Hi @neversi, can you write a test for this? Thanks!

neversi:
@edaniels it can be reproduced by tests with CGO_ENABLED=0

Also, I have added mock scaler to be able to pass tests for nocgo implementation and fix the ratio for i422 converter.

Now all tests pass without errors

edaniels (approved):
LGTM

PR #487, opened 2023-03-31
Add FPS Props for Linux Cameras

SeanAvery:
### Description
This PR allows for retrieving and setting FPS properties on Linux cameras. Like FrameSize, `Discrete` and `Stepwise` FrameRate returns are supported.

A simple check is added in the `fitness` function to avoid adding fps comparisons on MacOS and Screen drivers.

Implementation depends on this [blackjack/webcam PR](https://github.com/blackjack/webcam/pull/58).

Update: fps pr for blackjack/webcam has been merged!

#### Manual tests
Make sure there are no regressions across platforms (fitness distance is not affected).
- linux webcam
- mac webcam
- linux screen

#### Go tests
- Added unit tests for `float` and `FrameRate` matching in `prop_test.go`. Includes cases where FPS is not available.
- Added unit test for `calcFramerate` in `camera_linux_test.go`.
- No regressions in other prop.go and camera_linux tests.

### DIFF
--- a/pkg/driver/camera/camera_linux.go
+++ b/pkg/driver/camera/camera_linux.go
@@ -6,15 +6,17 @@ import "C"
 import (
 	"context"
 	"errors"
-	"github.com/pion/mediadevices/pkg/driver/availability"
 	"image"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/pion/mediadevices/pkg/driver/availability"
 
 	"github.com/blackjack/webcam"
 	"github.com/pion/mediadevices/pkg/driver"
@@ -314,13 +316,18 @@ func (c *camera) Properties() []prop.Media {
 			}
 
 			if frameSize.StepWidth == 0 || frameSize.StepHeight == 0 {
-				properties = append(properties, prop.Media{
-					Video: prop.Video{
-						Width:       int(frameSize.MaxWidth),
-						Height:      int(frameSize.MaxHeight),
-						FrameFormat: supportedFormat,
-					},
-				})
+				for _, framerate := range c.cam.GetSupportedFramerates(format, frameSize.MinWidth, frameSize.MinHeight) {
+					for _, fps := range enumFramerate(framerate) {
+						properties = append(properties, prop.Media{
+							Video: prop.Video{
+								Width:       int(frameSize.MaxWidth),
+								Height:      int(frameSize.MaxHeight),
+								FrameFormat: supportedFormat,
+								FrameRate:   fps,
+							},
+						})
+					}
+				}
 			} else {
 				// FIXME: we should probably use a custom data structure to capture all of the supported resolutions
 				for _, supportedResolution := range supportedResolutions {
@@ -339,13 +346,18 @@ func (c *camera) Properties() []prop.Media {
 						continue
 					}
 
-					properties = append(properties, prop.Media{
-						Video: prop.Video{
-							Width:       width,
-							Height:      height,
-							FrameFormat: supportedFormat,
-						},
-					})
+					for _, framerate := range c.cam.GetSupportedFramerates(format, uint32(width), uint32(height)) {
+						for _, fps := range enumFramerate(framerate) {
+							properties = append(properties, prop.Media{
+								Video: prop.Video{
+									Width:       width,
+									Height:      height,
+									FrameFormat: supportedFormat,
+									FrameRate:   fps,
+								},
+							})
+						}
+					}
 				}
 			}
 		}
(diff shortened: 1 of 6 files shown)


SeanAvery:
@edaniels @bazile-clyde Tagging for review.

edaniels (changes requested):
LGTM but can you write a few tests for this to make sure fitness is still working well 🏋️ ?

SeanAvery:
@edaniels Added tests.

pkg/driver/camera/camera_linux.go
```diff
@@ -385,3 +397,37 @@ func (c *camera) IsAvailable() (bool, error) {
...
+	}
+	return framerates
+}
+
+// calcFramerate turns fraction into a float32 fps value.
+func calcFramerate(numerator uint32, denominator uint32) (float32, error) {
+	if denominator == 0 {
+		return 0, errors.New("framerate denominator is zero")
+	}
+	// round to three decimal places to avoid floating point precision issues
```
bazile-clyde, line 431: Thanks for the comment here. Saved me a head scratch 😅

pkg/driver/camera/camera_linux.go
```diff
@@ -314,13 +316,18 @@ func (c *camera) Properties() []prop.Media {
...
+								Height:      int(frameSize.MaxHeight),
+								FrameFormat: supportedFormat,
+								FrameRate:   fps,
+							},
+						})
+					}
+				}
 			} else {
 				// FIXME: we should probably use a custom data structure to capture all of the supported resolutions
 				for _, supportedResolution := range supportedResolutions {
```
bazile-clyde, line 333: I know you didn't write this code but just curious if you know why this lib would limit the resolutions to the ones in `supportedResolutions` instead of allowing everything the camera returns. It isn't obvious to me at all.

SeanAvery: Yes I think we can safely handle stepwise `FrameSize` returns without hardcoded list. 

Here is the [PR](https://github.com/pion/mediadevices/commit/16bcd0b7dd6932e2f626c6bcf05192480e908978) that added list.

This is also relevant to stepwise FrameRate returns - may need to revisit the data-structures for cleaner stepwise handling.

bazile-clyde: Ah, yeah I totally misread the FIXME. Looks like someone planned on doing exactly that. Probably isn't a big deal either for now.

pkg/prop/prop.go
```diff
@@ -145,11 +145,10 @@ func (p *MediaConstraints) FitnessDistance(o Media) (float64, bool) {
 	cmps.add(p.Width, o.Width)
 	cmps.add(p.Height, o.Height)
 	cmps.add(p.FrameFormat, o.FrameFormat)
-	// The next line is comment out for now to not include framerate in the fitness function.
-	// As camera.Properties does not have access to the list of available framerate at the moment,
-	// no driver can be matched with a framerate constraint.
-	// Note this also affect screen caputre as screen.Properties does not fill in the Framerate field.
-	// cmps.add(p.FrameRate, o.FrameRate)
+	// skip framerate if not available in media properties
+	if o.FrameRate > 0.0 {
```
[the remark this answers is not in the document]
bazile-clyde: What are the consequences of relying the logic in the second bullet? 

I'm guessing for Mac OS X, since the framerate isn't currently supported in this lib everything will be penalized by the same amount and we'll still prioritize drivers the same? 

For Linux it seems like requesting a framerate and for some reason not being able to retrieve one would penalize the props just as much as having a wrong framerate which seems fine.

To do the first bullet you'd have to do something like...

```
type FrameRateConstraint interface {
	Compare(frame.Format) (float64, bool)
	Value() (frame.Format, bool)
}

type Float float32

func (f Float) Compare(a float32) (float64, bool) {
	if a < 0.0 {
		return 0.0, true
	}
	return FloatExact(float32(f)).Compare(a)
}
...

```
Which isn't too bad but might be overkill if the `FloatConstraint` logic works as is.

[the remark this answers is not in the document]
SeanAvery: 1. FrameRateConstraint

This seems like a decent option. The downside is increasing the surface area of prop code. The implementation would be the exact same as `FloatConstraint` with unset checks, which seems redundant. Perhaps there is a nice golang way to wrap FloatConstraint?

It seems much simpler to add a one-liner check in FitnessDistance.

2. Allow unset FPS props

`selectBestDriver` would still function properly. The only potential issue would be for users who use `MediaConstraints.FitnessDistance` directly as there would be a breaking change to fitness scores. 

I am assuming this is a rare case, as most users are going through the `GetUserMedia`/`selectVideo` endpoints. Also scores would only be affected if FrameRate is specified in the `MediaStreamConstraints.`

~Note: Only relavant when FrameRate is

SeanAvery:
Cleaning up git history in preparation for merge.

PR #493, opened 2023-04-11
Add default Huffman table if huffman table uninitialized

kim-mishra:
#### Description
Certain MJPEG frames do not include Huffman tables, this causes an `uninitialized Huffman table` error during `jpeg.Decode()`. These changes add a default Huffman Table in the case that one was not provide, so that the jpeg can be processed and transformed into an image. I tested these changes locally with cameras that provide and do not provide Huffman tables to ensure that the error is avoided, and for frames with initialized Huffman tables there is no change in behavior.

#### Reference issue
Fixes an uninitialized Huffman error that is returned from `jpeg.Decode()` when a jpeg Frame omits the Huffman Table


pkg/frame/compressed.go
```diff
@@ -6,7 +6,25 @@ import (
...
+	sosMarker = []byte{255, 218}
+)
+
 func decodeMJPEG(frame []byte, width, height int) (image.Image, func(), error) {
 	img, err := jpeg.Decode(bytes.NewReader(frame))
+
+	if err != nil && err.Error() == "invalid JPEG format: uninitialized Huffman table" {
```
edaniels, line 20: can you return early if there is no error?
also, it would be better to use errors.As with https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/image/jpeg/reader.go;l=18 to then check just the "uninitialized Huffman table" error so we can be a bit more specific

pkg/frame/compressed.go
```diff
@@ -6,7 +6,25 @@ import (
 	"image/jpeg"
 )
 
+// Thank you to https://github.com/filiptc/gorbit/blob/fa87ff39b68a6706306f34c318e0b9a5a3c97110/image/overlay.go#L37-L40 for addMotionDht, dhtMarker, dht, and sosMarker
```
edaniels, line 9: licensing checks out 👍

pkg/frame/compressed.go
```diff
@@ -6,7 +6,25 @@ import (
...
+	if err != nil && err.Error() == "invalid JPEG format: uninitialized Huffman table" {
+		img, err = jpeg.Decode(bytes.NewReader(addMotionDht(frame)))
+	}
+
 	return img, func() {}, err
 }
+
+func addMotionDht(frame []byte) []byte {
+	jpegParts := bytes.Split(frame, sosMarker)
+	return append(jpegParts[0], append(dhtMarker, append(dht, append(sosMarker, jpegParts[1]...)...)...)...)
```
edaniels, line 29: this should check that the split worked. otherwise a malicious input could crash the caller. this would be a good test

edaniels, line 29: this is copying bytes in a potentially expensive way. can you make the slice in advance with the new size and then copy in order. it will also make these appends a tad easier to read!

pkg/frame/compressed.go
edaniels: we do but we can eliminate the first part of that string to check

pkg/frame/compressed.go
```diff
@@ -6,7 +6,55 @@ import (
...
... 106, 115, 116, 117, 118, 119, 120, 121, 122, 131, 132, 133, 134, 135, 136, 137, 138, 146, 147, 148, 149, 150, 151, 152, 153, 154, 162, 163, 164, 165, 166, 167, 168, 169, 170, 178, 179, 180, 181, 182, 183, 184, 185, 186, 194, 195, 196, 197, 198, 199, 200, 201, 202, 210, 211, 212, 213, 214, 215, 216, 217, 218, 225, 226, 227, 228, 229, 230, 231, 232, 233, 234, 241, 242, 243, 244, 245, 246, 247, 248, 249, 250, 17, 0, 2, 1, 2, 4, 4, 3, 4, 7, 5, 4, 4, 0, 1, 2, 119, 0, 1, 2, 3, 17, 4, 5, 33, 49, 6, 18, 65, 81, 7, 97, 113, 19, 34, 50, 129, 8, 20, 66, 145, 161, 177, 193, 9, 35, 51, 82, 240, 21, 98, 114, 209, 10, 22, 36, 52, 225, 37, 241, 23, 24, 25, 26, 38, 39, 40, 41, 42, 53, 54, 55, 56, 57, 58, 67, 68, 69, 70, 71, 72, 73, 74, 83, 84, 85, 86, 87, 88, 89, 90, 99, 100, 101, 102, 103, 104, 105, 106, 115, 116, 117, 118, 119, 120, 121, 122, 130, 131, 132, 133, 134, 135, 136, 137, 138, 146, 147, 148, 149, 150, 151, 152, 153, 154, 162, 163, 164, 165, 166, 167, 168, 169, 170, 178, 179, 180, 181, 182, 183, 184, 185, 186, 194, 195, 196, 197, 198, 199, 200, 201, 202, 210, 211, 212, 213, 214, 215, 216, 217, 218, 226, 227, 228, 229, 230, 231, 232, 233, 234, 242, 243, 244, 245, 246, 247, 248, 249, 250}
```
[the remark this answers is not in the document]
kim-mishra: This should work across devices in the case that the Huffman encoding info has been omitted for bandwidth optimization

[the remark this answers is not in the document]
edaniels: Can you point where this comes from though?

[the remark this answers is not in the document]
kim-mishra: The [JPEG wiki](https://en.wikipedia.org/wiki/JPEG) lists `dhtMarker` & `sosMarker` as dht & sos respectively.
This [RFC](https://www.rfc-editor.org/rfc/rfc2435) discusses a "Standard Huffman Table" and location of the citation is behind a paywall. This [thread](https://github.com/blackjack/webcam/issues/7) is where I got the table from. It works on cameras I have tested with and has been reported to work on other cameras as well.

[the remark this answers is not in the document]
at-wat: > This [RFC](https://www.rfc-editor.org/rfc/rfc2435) discusses a "Standard Huffman Table" and location of the citation is behind a paywall.

https://www.w3.org/Graphics/JPEG/itu-t81.pdf
W3C seems putting it in public.

pkg/frame/compressed.go
at-wat: Strictly speaking, when OSS licensed code is copied, we should include the full original license statements according to
```
The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.
```
in case of MIT.
https://github.com/filiptc/gorbit/blob/fa87ff39b68a6706306f34c318e0b9a5a3c97110/LICENSE

pkg/frame/compressed.go
```diff
@@ -2,11 +2,89 @@ package frame
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+*/
```
kim-mishra, line 34: @at-wat like this?

edaniels (changes requested):
Done with another pass. Thanks for adding tests!

pkg/frame/compressed.go
```diff
@@ -2,11 +2,89 @@ package frame
...
 func decodeMJPEG(frame []byte, width, height int) (image.Image, func(), error) {
 	img, err := jpeg.Decode(bytes.NewReader(frame))
+	if err == nil {
+		return img, func() {}, err
+	}
+
+	var target jpeg.FormatError = jpeg.FormatError("uninitialized Huffman table")
+	if errors.As(err, &target) {
+		if err.Error() == target.Error() {
+			img, err = jpeg.Decode(bytes.NewReader(addMotionDht(frame)))
```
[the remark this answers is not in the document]
kim-mishra: Could that break other things? The function did not originally ever return nil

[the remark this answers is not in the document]
edaniels: it would not

pkg/frame/compressed.go
```diff
@@ -2,11 +2,84 @@ package frame
...
+
+func addMotionDht(frame []byte) []byte {
+	jpegParts := bytes.Split(frame, sosMarker)
+	if len(jpegParts) != 2 {
+		return frame
+	}
+	correctedFrame := make([]byte, len(jpegParts[0])+huffmanTableInfoLength+len(jpegParts[1]))
+	correctedFrameOffset := 0
+
+	copy(correctedFrame[correctedFrameOffset:], jpegParts[0])
```
edaniels, line 71: so much better!

pkg/frame/compressed_test.go
```diff
@@ -0,0 +1,52 @@
...
+	expectedErrorMessage := "invalid JPEG format: uninitialized Huffman table"
+	if err.Error() != expectedErrorMessage {
+		t.Errorf("Wrong decode error result,\nexpected:\n%+v\ngot:\n%+v", expectedErrorMessage, err)
+		t.FailNow()
+	}
+
+	// Decode passes after adding default Huffman table to
+	defaultHuffmanTableFrame, err := jpeg.Decode(bytes.NewReader(addMotionDht(UninitializedHuffmanTable)))
+	if err != nil {
+		t.Errorf("Expected decode function to pass after adding default Huffman table. Failed with %v\n", err)
```
edaniels, line 22: all of these need the failnow too but you can just do Fatalf instead

edaniels (changes requested):

pkg/frame/compressed_test.go
```diff
@@ -0,0 +1,51 @@
...
+		t.Fatalf("addMotionDht updated the frame despite being improperly constructed")
+	}
+}
+
+func TestDecodeMJPEG(t *testing.T) {
+	_, _, err := decodeMJPEG(UninitializedHuffmanTable, 640, 480)
+	if err != nil {
+		t.Fatalf("Expected decode function to pass. Failed with %v\n", err)
+	}
+}
```
kim-mishra, line 51: @edaniels like this?

edaniels: yep!

edaniels (approved):
LGTM. Will wait on @at-wat for other approval

pkg/frame/compressed.go
at-wat: Maybe better to move the copied part to separated file (if code inside `decodeMJPEG` and `addMotionDht` is also copied, current structure is fine)
Could you add a URL of the original repo as well?

Other things looks good to me!

kim-mishra: Link is added! parts of the functions you mentioned are taken from the repo

at-wat (approved):
LGTM!

PR #494, opened 2023-04-12
expose device name and bus info

bazile-clyde:
#### Description
This PR exposes the device name and bus info for Linux cameras.


pkg/driver/camera/camera_linux.go
```diff
@@ -112,7 +112,22 @@ func discover(discovered map[string]struct{}, pattern string) {
 		if reallink == prioritizedDevice {
 			priority = driver.PriorityHigh
 		}
+
+		name := ""
+		busInfo := ""
```
edaniels, line 117:
```suggestion
		var name, busInfo string
```

edaniels (approved):
LGTM with a nit

PR #495, opened 2023-04-13
allow adapters to handle their own state

bazile-clyde:
#### Description
This PR allows for adapters (video and audio drivers) to handle their own state. This allows us to fix a few issues

- Calling one of the `Initialize` functions resets the state of an adapter to `StateClosed` even if it's in use. 
- Disconnected drivers will return a stale state, possibly `StateClosed` or `StateRunning`.
- Drivers used by other applications will start with a state of `StateClosed`.

Ideally, all drivers would return the actual state of the driver on queries. This PR returns real states for `camera_linux.go` fixing the above issues. All other drivers keep their current behavior. 

#### Testing
This was manually tested on a Linux device and works as expected. It should not break any tests or change the current behavior for any drivers other than those on Linux.

#### Advice for Reviewers 
The easiest way to review this code is to start with the changes in `driver.go` and `wrapper.go`. Afterwards, ensure none of the logic in any of the files besides `camera_linux.go` has changed. Lastly, review/test the changes in `camera_linux.go` and the new states in `state.go`. Simply running a background goroutine that continuously calls and prints the `Status()` of a Linux camera and observing what happens when the camera is disconnected and reconnected should be enough.


pkg/driver/wrapper_test.go
```diff
@@ -17,52 +17,21 @@ type adapterMock struct{}
...
+func (a *videoAdapterBrokenMock) Status() State { return StateClosed }
 
 type audioAdapterBrokenMock struct{ adapterMock }
 
-func (a *audioAdapterBrokenMock) AudioRecord(p prop.Media) (r audio.Reader, err error) {
+func (a *audioAdapterBrokenMock) AudioRecord(_ prop.Media) (r audio.Reader, err error) {
 	return nil, recordErr
 }
-
-func TestVideoWrapperState(t *testing.T) {
```
bazile-clyde, line 42: This test and the one removed below can no longer test state errors by calling mock functions since that functionality has been pushed down to the level of the adapter. This means state information can only be obtained by querying real drivers, not mocks.

pkg/driver/audiotest/dummy.go
```diff
@@ -15,28 +15,36 @@ import (
...
+		d.closed = ctx.Done()
+		d.cancel = cancel
+		return nil
+	})
 }
 
 func (d *dummy) Close() error {
-	d.cancel()
-	return nil
+	return d.state.Update(driver.StateClosed, func() error {
```
edaniels, line 38: it seems like for each of these closed states now, that if closing fails, the camera is not closed and we are an indeterminate state

pkg/driver/camera/camera_linux.go
```diff
@@ -303,6 +304,10 @@ func (c *camera) VideoRecord(p prop.Media) (video.Reader, error) {
 }
 
 func (c *camera) Properties() []prop.Media {
+	if c.cam == nil {
```
edaniels, line 307: why would the camera be nil?

pkg/driver/camera/camera_linux.go
```diff
@@ -350,3 +355,28 @@ func (c *camera) Properties() []prop.Media {
 	}
 	return properties
 }
+
+func (c *camera) Status() driver.State {
+	var index int32
```
edaniels, line 360: index can move down into the if along with camera

pkg/driver/videotest/dummy.go
```diff
@@ -129,10 +137,10 @@ func (d *dummy) VideoRecord(p prop.Media) (video.Reader, error) {
 		}, func() {}, nil
 	})
 
-	return r, nil
+	return r, d.state.Update(driver.StateRunning, func() error { return nil })
 }
 
-func (d dummy) Properties() []prop.Media {
+func (d *dummy) Properties() []prop.Media {
```
edaniels, line 143: why did this change to being a pointer?

pkg/driver/videotest/dummy.go
```diff
@@ -144,3 +152,7 @@ func (d dummy) Properties() []prop.Media {
 		},
 	}
 }
+
+func (d *dummy) Status() driver.State {
+	return d.state
```
edaniels, line 157: should these states be atomic values?

pkg/driver/driver.go
```diff
@@ -37,11 +37,11 @@ type Adapter interface {
...
 	Close() error
 	Properties() []prop.Media
+	Status() State
 }
 
 type Driver interface {
 	Adapter
 	ID() string
 	Info() Info
-	Status() State
```
edaniels, line 46: did this need to move? if so, why?

pkg/driver/state.go
```diff
@@ -16,24 +16,34 @@ const (
 	// StateRunning means that the driver has been sending data. The caller
 	// who started the driver may start reading data from the hardware.
 	StateRunning = "running"
+	// StateGood means the driver is available for use by applications.
```
edaniels, line 19: you mentioned that only linux returns this now but this means that end users need to understand all of these states. is there a way to use all of the states or merge the new ones together? I'm concerned an end user may not know how to interpret the state.

edaniels (changes requested):
Nice changes! Done with the first pass here. I'd like to see some tests added as well as manual testing performed on windows and macos.

pkg/driver/state.go
```diff
@@ -16,24 +16,34 @@ const (
...
+	StateGood = "good"
```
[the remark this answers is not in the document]
bazile-clyde: Because this is now dependent on a specific driver it's a lot more difficult to test. For Linux devices this requires an ioctls call. Short of mocking the return values for ioctls or setting up a virtual v4l2 device I don't think there's a good way to test it. This is also a public function that isn't needed much internally so you wouldn't be able to mock out the `Status` function and see the behavior of internal functions change either. It's tricky.

pkg/driver/camera/camera_linux.go
bazile-clyde: so `c.cam` is set in the `Open` method but devices are registered in the `discover` function. That leaves a window when you can call `Properties` before you call `Open` on a device which will crash your program. I've hit this case while testing.

pkg/driver/state.go
```diff
@@ -16,24 +16,34 @@ const (
...
+	StateBad = "bad"
+	// StateNotFound means the driver cannot be found or is disconnected.
+	StateNotFound = "not found"
 )
 
-// Update updates current state, s, to next. If f fails to execute,
-// s will stay unchanged. Otherwise, s will be updated to next
-func (s *State) Update(next State, f func() error) error {
+// CheckUpdate returns nil if the current state can be updated to next state. Otherwise, it returns an error.
+func (s *State) CheckUpdate(next State) error {
```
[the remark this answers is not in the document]
bazile-clyde: That's true. The new statuses don't cleanly map to the state-machine-like states we've had before. For example, you can open v4l2 drivers twice without issue but the this function will error if you do. I added an error for trying to use the `CheckUpdate` function with any of the new states. Ideally, this function would go away after we've updated the other drivers but for now it's only used for legacy behavior.

pkg/driver/videotest/dummy.go
bazile-clyde: auto fix by IDE for mixed value and pointer receivers. Changed back.

pkg/driver/videotest/dummy.go
bazile-clyde: It wasn't before so I didn't change it to be. The `adapterWrapper` class just returns and modifies its values without any mutexes so pushing down this logic won't cause any new issues. It seems like the `adapterWrapper` class in general should be protected by mutexes. I think that would be out of scope for this change though.

edaniels:
@at-wat could you take a second set of eyes on this?

pkg/driver/state.go
edaniels: Got it. Let's see if we can work with this to keep the existing enums.
1. This would be StateClosed; nothing in the process has opened it yet
2. This would be StateRunning as you already pointed out
3. This would be StateClosed if we consider closed to be the default
4. This would be StateClosed if we consider that if we cannot find something, it is not open

If we did things this way, could the status of camera linux use its current state plus these cases to produce the right value?

bazile-clyde: I don't think so given the ambiguity of open and closed for a v4l2 driver _but_ I think there's a better way. Since the big thing we want to avoid here is breaking code for current users would versioning these changes work?

These changes are to support exposing the experimental `Initialize` function since calling that function ruins the state logic we have in place. For current users, neither this change nor the `Initialize()` function is needed. We could put these changes in `pion/mediadevices/tree/master/pkg/experimental` or a `v2` folder and merge them all at once in a future version once they're more stable. I _would_ suggest adding these changes to a different major version but since we don't have one yet, and as far as I've seen there is no current consensus on what the first version of this software would look like, it's kind of a moot point.

edaniels: Why wouldn't that work given the ambiguity? What breaks upstream? I' m not sure about adding a v2/experimental for this if we can still make something work

edaniels: You do raise a good point that this isn't 1.0 yet but I'm still curious if we can make this work with what we have an addition to @at-wat's feedback

bazile-clyde: > Why wouldn't that work given the ambiguity?

What exactly does an _open_ and _closed_ state mean or imply for a device driver? In the truest sense it means something has an open file descriptor to it. It implies nothing in particular. For example, if the user has a device at `/dev/video0` we can see what processes opened this file 
```
$ lsof -t /dev/video0
46915
...
$ ps -ef | grep myApp  
$USER      46915   46729  7 15:14 pts/4    00:00:18 myApp
```  
If it's _closed_ then `lsof` will have an empty process list and no output. What does this imply or in other words, what can (or can't) we do based off of this information? To get the status of the driver we need to open the driver, query its status, then close the driver. The driver may or may not have already been opened or closed. If we continue to assume that drivers can only be opened once (see [toOpened](https://github.com/pion/mediadevices/blob/bf290b026c1037ba0ff839a41d42d86f666d4455/pkg/driver/state.go#L45)) we will never be able to get the status of an opened driver because we won't be able to open it. 

>  1. This would be StateClosed; nothing in the process has opened it yet

We call `Initialize` and discover what we think is a new driver. Have we opened it yet? We would need to map the `camera.path` to a boolean indicating whether we called `Open`. This ignores the fact that another process may have opened the file and I'm not sure who this info is useful for per my comment above.

> 3. This would be StateClosed if we consider closed to be the default

There's not a "default" state for a driver that you can't get state information from. See some common error states [here](https://www.kernel.org/doc/html/v4.9/media/uapi/gen-errors.html#id1). Those errors are returned from the (VIDIOC_LOG_STATUS)[https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/vidioc-log-status.html] ioctl. Device drivers aren't guaranteed to implement that ioctl so I used a different set of ioctls but same deal. 

> 4. This would be StateClosed if we consider that if we cannot find something, it is not open

How would a user know if they can attempt to open a driver and start streaming? If both condition 4 and 1 are the same it could mean either, "nothing is streaming from this driver so feel free to do so" or "this driver is not connected so do not stream." It's inherently ambiguous. 

 > What breaks upstream? 
 
 Anyone using the `driver.Status` will now have to update their code based on these new states. It doesn't necessarily break the API in the literal sense but users may have to re-write code or this will introduce subtle runtime errors.

pkg/driver/audiotest/dummy.go
```diff
@@ -81,7 +89,7 @@ func (d *dummy) AudioRecord(p prop.Media) (audio.Reader, error) {
 		}
 		return a, func() {}, nil
 	})
-	return reader, nil
+	return reader, d.state.Update(driver.StateRunning, func() error { return nil })
```
at-wat, line 92: Is it safe to allow changing state between `CheckUpdate` and `Update`, then return `invalid state` error?
Maybe it doesn't happen in typical usages?

pkg/driver/state.go
```diff
@@ -16,24 +16,39 @@ const (
...
+	// This function should not be used with the states below. This function is kept solely to support the legacy states.
+	if next == StateAvailable || next == StateUnavailable || next == StateDisconnected {
+		return fmt.Errorf("invalid state: cannot update to state '%s'", string(next))
```
at-wat, line 31: Better to expose base `invalid state` error and wrap by `%w`

at-wat (commented):
The main changes looks nice!

pkg/driver/state.go
edaniels: Noodling on a thought. Considering windows and macos won't be updated here. Maybe a temporary pre 1.0 thing could be to introduce an adapter specific state that can be optionally used in conjunction with driver state. Then v1 can merge the two or have your way supersede the two

bazile-clyde:
@edaniels @at-wat Thanks for the reviews. After seeing all of the comments and speaking with @edaniels I updated this PR to make this functionality optional. It's far less changes now and will not break the API.

pkg/driver/availability/error.go
```diff
@@ -0,0 +1,25 @@
+package availability
+
+import (
+	"errors"
+)
+
+type Error interface {
+	Error() string
```
edaniels, line 8: You'll need to add a unexported method of some bogus name to this otherwise this will be implemented by any error

pkg/driver/availability/error.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+var ErrUnimplemented Error = errors.New("not implemented")
```
edaniels, line 11: Can put this in a var block

pkg/driver/availability/error.go
edaniels: Alternatively you can use a no interface approach for errors like this. Look at the must rebuild error in RDK if you're interested

edaniels (approved):
Nice. I think this is actually really clean now. Thanks for discussing this. I have some comments around the error itself but they're fairly nitpicky. If you disagree with them, I have no problem with this merging in.

pkg/driver/availability/error.go
bazile-clyde: Done. Used the no interface approach. Works nicely.

PR #502, opened 2023-05-11
Update README.md to include mention of GetDisplayMedia

infamy:
Added to the FAQ to mention GetDisplayMedia if using driver/screen in the section explaining solutions to the error that gets throne if you don't make that swap in example code.

#### Reference issue
Fixes #...


at-wat (approved):
LGTM

PR #507, opened 2023-06-02
Add YUYV to supported video formats

RyanBoring:
#### Description
Add YUYV to supported video formats; used to be implicitly supported under YUY2. That meant the user had to know to ask for YUY2 decoding when using a YUYV device; this PR allows the user to specify YUYV instead.

### DIFF
--- a/pkg/frame/decode.go
+++ b/pkg/frame/decode.go
@@ -7,25 +7,26 @@ import (
 type Format string
 
 const (
-	// FormatI420 https://www.fourcc.org/pixel-format/yuv-i420/
+	// FormatI420 https://wiki.videolan.org/YUV#I420
 	FormatI420 Format = "I420"
 	// FormatI444 is a YUV format without sub-sampling
 	FormatI444 Format = "I444"
-	// FormatNV21 https://www.fourcc.org/pixel-format/yuv-nv21/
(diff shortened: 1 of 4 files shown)


RyanBoring:
@bazile-clyde Would you be able to review this for me? I can't add you as a reviewer directly

bazile-clyde:
> @bazile-clyde Would you be able to review this for me? I can't add you as a reviewer directly

@RyanBoring yep! Looking now.

pkg/frame/decode.go
```diff
@@ -17,6 +17,8 @@ const (
 	FormatNV12 = "NV12"
 	// FormatYUY2 https://www.fourcc.org/pixel-format/yuv-yuy2/
 	FormatYUY2 = "YUY2"
+	// FormatYUYV https://www.fourcc.org/pixel-format/yuv-yuy2/
```
bazile-clyde, line 20: This link isn't working for me. I think it's gone stale. Maybe figure our if _fourcc.org_ has moved to another domain, and if so where. Or you could paste another authoritative link about the YUYV format.

RyanBoring: Wow yeah great catch, I just changed them in a new commit

bazile-clyde (approved):
Nice! LGTM!

pkg/frame/decode.go
bazile-clyde: Thank you!

bazile-clyde (approved):

PR #509, opened 2023-06-11
Added Name in driver.Info to Microphone driver

aljanabim:
A tiny change to the Initialize function in the Microphone driver ensuring that the device name is included in the driver.Info struct.


at-wat (approved):
LGTM!

PR #513, opened 2023-07-04
Fix invalid memory access in ScalerFastBoxSampling

at-wat:
`[src-px]->[dst-px]` of 4x1px to 2x1px scaling was
```
0->0, 1->1, 2->1, 3->2(out of bounds)
```

Fix it to be
```
0->0, 1->0, 2->1, 3->1
```


edaniels (approved):
LGTM

PR #551, opened 2024-01-23
Update malgo to v0.11.11-0.20231225022531-dd505cb7efd5 for a macOS microphone fix

edaniels:
- Also log when we do not support a malgo format (e.g. the iota value)


at-wat (approved):
LGTM

PR #556, opened 2024-02-28
Add fix that strips unexpected padding for NV12 frames

hexbabe:
On macOS Sonoma, using the Logitech C922 Pro webcam, NV12 frames seem to come with a default min amount of padding for smaller resolutions e.g. 160x90. This would result in smaller resolution NV12 images outputting like this:

videoframe_16623 my upside down forehead 

This change crops the data—making sure it is packed tightly—so that the `decodeNV12` precondition can be preserved.


edaniels (approved):
LGTM. Thanks for this fix!

PR #568, opened 2024-04-26
Add webcam close to linux camera discover

seanavery:
#### Description

This PR fixes an issue with dangling open files from Linux camera Initialize/discover calls (we are missing a webcam close).


edaniels (approved):
LGTM

Issue #347, opened 2021-07-14
Windows 10 amd64 can not open Camera


fjfo:
### Your environment.
  - Version: golang 1.16
  - OS: Windows 10, amd64
### What did you do?
        x264Params, err := x264.NewParams()
	must(err)
	x264Params.Preset = x264.PresetMedium
	x264Params.BitRate = 1_000_000 // 1mbps

	codecSelector := mediadevices.NewCodecSelector(
		mediadevices.WithVideoEncoders(&x264Params),
	)

	mediaStream, err := mediadevices.GetUserMedia(mediadevices.MediaStreamConstraints{
		Video: func(c *mediadevices.MediaTrackConstraints) {
			c.FrameFormat = prop.FrameFormat(frame.FormatI420)
			c.Width = prop.Int(640)
			c.Height = prop.Int(480)
		},
		Codec: codecSelector,
	})
	must(err)

	videoTrack := mediaStream.GetVideoTracks()[0].(*mediadevices.VideoTrack)
	defer videoTrack.Close()
	videoReader := videoTrack.NewReader(false)
	var buf bytes.Buffer

		imgFrame, release, err := videoReader.Read()
		if err == io.EOF {
			return
		}
		must(err)

		out, _ := os.Create("./img.jpeg")
		defer out.Close()

		err = jpeg.Encode(out, imgFrame, nil)
		if err != nil {
			log.Println(err)
		}
		release()
		buf.Reset()

### What did you expect?
- Open Camera and write 1 Frame to disk
### What happened?
- Error: errConnectFilters = "failed to connect filters";

fjfo:
More debug log here

`# github.com/pion/mediadevices/pkg/driver/camera
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: $WORK\b184\_x002.o: in function `_cgo_a5197d6edf98_Cfunc_getName':
/tmp/go-build/cgo-gcc-prolog:119: undefined reference to `getName'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: $WORK\b184\_x002.o: in function `_cgo_a5197d6edf98_Cfunc_getProp':
/tmp/go-build/cgo-gcc-prolog:139: undefined reference to `getProp'
collect2.exe: error: ld returned 1 exit status`

[...]

PR #569, opened 2024-04-28
Remove Windows camera cpp inline function 

qiulin:
Description
Reference issue
Parttial fixes https://github.com/pion/mediadevices/issues/347, remove inline function in cpp, which will cause compiling problem when debug


qiulin:
@edaniels Hi, Can you review the code? Thanks.

pkg/driver/camera/camera_windows.cpp
```diff
@@ -347,6 +359,7 @@ int openCamera(camera* cam, const char** errstr)
   if (src == nullptr || dst == nullptr ||
       FAILED(hr = graphBuilder->Connect(src, dst)))
   {
+    printErr(hr);
```
edaniels, line 362: does the error need to be printed here?

pkg/driver/camera/camera_windows.go
```diff
@@ -37,7 +37,7 @@ func init() {
 
 // Initialize finds and registers camera devices. This is part of an experimental API.
 func Initialize() {
-	C.CoInitializeEx(nil, C.COINIT_MULTITHREADED)
+	C.CoInitializeEx(nil, C.COINIT_APARTMENTTHREADED)
```
edaniels, line 40: What is this change?

pkg/driver/camera/camera_windows.cpp
qiulin: It's just printing error for debug, I will resolve it

pkg/driver/camera/camera_windows.go
qiulin: It's recommended to work with some UI framework, like Wails (https://wails.io/).

qiulin: I reverted it, and will open another PR.

qiulin: My English is bad, please forgive any mistakes.

edaniels (approved):
LGTM

PR #571, opened 2024-05-01
Bump webcam dependency

seanavery:
#### Description

Simple webcam bump to `0.6.1` -- fixes issue with mmap flags causing streams to fail for some Linux drivers.
See webcam commit [here](https://github.com/blackjack/webcam/commit/15347780752e4f6e9767d4e8b15a60038dc27f85).


edaniels:
Yeah it's fine for now. I may decommission them

PR #579, opened 2024-08-27
add libopus-android-armv7 to opus

snowlyg:
add libopus-android-armv7 to opus


at-wat (commented):
Thank you for your contribution!

Please update https://github.com/pion/mediadevices/tree/master/dockerfiles and https://github.com/pion/mediadevices/blob/master/Makefile to make library building process reproducible.

snowlyg:
> Thank you for your contribution!
> 
> Please update https://github.com/pion/mediadevices/tree/master/dockerfiles and https://github.com/pion/mediadevices/blob/master/Makefile to make library building process reproducible.

This dockerfile is only useful to opus because openh264 must need NDK to build.

snowlyg:
@at-wat I'm used v4 webrtc and update  v3 to v4 change too much files. so i close it.

PR #611, opened 2025-03-07
Feat/add helper function for av1 and h265 codec

3DRX:
I'm working on a ffmpeg based encoder (using https://github.com/asticode/go-astiav) module that supports h264_nvenc, h265_nvenc, and av1_nvenc, as I tried to use AV1 codec, I discovered that there is no helper function for creating an AV1 codec in this library for now, so I need to create one in my project's repo. But this should be in pkg/codec/codec.go in the future.

- [x] For this PR to be merged, #608 needs to be merged first since the updated pion/webrtc v4.0.12 contains a refactored AV1 RTP packetizer.


3DRX:
Update: just added `NewRTPH265Codec` as well, both `NewRTPH265Codec` and `NewRTPAV1Codec` are tested in my project.

kevmo314 (approved):

PR #612, opened 2025-03-07
Add EncoderController method to VideoTrack and AudioTrack.

3DRX:
## Why?

Currently, encoders in mediadevices can implement `func SetBitrate(bitrate int64) error`. But when the mediadevices API is used with pion/webrtc and congestion controller like the following (which is a common usage), it's not straight forward to call SetBitrate in the estimator.OnTargetBitrateChange callback.

```go
	codecselector := mediadevices.NewCodecSelector(
		mediadevices.WithVideoEncoders(&params),
	)
	m := &webrtc.MediaEngine{}
	codecselector.Populate(m)
	i := &interceptor.Registry{}
	if err := webrtc.RegisterDefaultInterceptors(m, i); err != nil {
		panic(err)
	}
	congestionControllerFactory, err := cc.NewInterceptor(func() (cc.BandwidthEstimator, error) {
		return gcc.NewSendSideBWE(gcc.SendSideBWEInitialBitrate(400_000))
	})
	if err != nil {
		panic(err)
	}
	estimatorChan := make(chan cc.BandwidthEstimator, 1)
	congestionControllerFactory.OnNewPeerConnection(func(id string, estimator cc.BandwidthEstimator) { //nolint: revive
		estimatorChan <- estimator
	})
	i.Add(congestionControllerFactory)
	if err := webrtc.ConfigureTWCCHeaderExtensionSender(m, i); err != nil {
		panic(err)
	}
	if err := webrtc.ConfigureCongestionControlFeedback(m, i); err != nil {
		panic(err)
	}
	api := webrtc.NewAPI(webrtc.WithMediaEngine(m), webrtc.WithInterceptorRegistry(i))
	config := webrtc.Configuration{
		ICEServers: []webrtc.ICEServer{
			{
				URLs: []string{"stun:stun.l.google.com:19302"},
			},
		},
	}
	peerConnection, err := api.NewPeerConnection(config)
	if err != nil {
		panic(err)
	}

	mediaStream, err := mediadevices.GetUserMedia(mediadevices.MediaStreamConstraints{
		Video: func(constraint *mediadevices.MediaTrackConstraints) {
			constraint.Width = prop.Int(1920)
			constraint.Height = prop.Int(1080)
			constraint.FrameRate = prop.Float(90)
		},
		Codec: codecselector,
	})
	if err != nil {
		panic(err)
	}
	for _, videoTrack := range mediaStream.GetVideoTracks() {
		_, err := peerConnection.AddTransceiverFromTrack(
			videoTrack,
			webrtc.RTPTransceiverInit{
				Direction: webrtc.RTPTransceiverDirectionSendonly,
			},
		)
		if err != nil {
			panic(err)
		}
		break // assume only one video track
	}

	estimator := <-estimatorChan

	estimator.OnTargetBitrateChange(func(bitrate int) {
		// there is no way to propagate the updated bitrate to encoder, since we can't get the encoder controller of the track, we can't call SetBitrate here
	})
```

To address this problem, we need to create a new function to be called by user to get the `codec.EncoderController` object from track, after `track.EncoderController` implemented, the problem in above code can be solved as

```go
	// ... (same)

	mediaStream, err := mediadevices.GetUserMedia(mediadevices.MediaStreamConstraints{
		Video: func(constraint *mediadevices.MediaTrackConstraints) {
			constraint.Width = prop.Int(1920)
			constraint.Height = prop.Int(1080)
			constraint.FrameRate = prop.Float(90)
		},
		Codec: codecselector,
	})
	if err != nil {
		panic(err)
	}
	var encoderController codec.EncoderController
	for _, videoTrack := range mediaStream.GetVideoTracks() {
		_, err := peerConnection.AddTransceiverFromTrack(
			videoTrack,
			webrtc.RTPTransceiverInit{
				Direction: webrtc.RTPTransceiverDirectionSendonly,
			},
		)
		if err != nil {
			panic(err)
		}
		encoderController = videoTrack.EncoderController()
		break // assume only one video track
	}

	estimator := <-estimatorChan

	estimator.OnTargetBitrateChange(func(bitrate int) {
		// we can call SetBitrate from encoderController
		if encoderController == nil {
			return
		}
		bitrateController, ok := encoderController.(codec.BitRateController)
		if !ok {
			return
		}
		bitrateController.SetBitRate(bitrate)
	})
```


track.go
```diff
@@ -449,6 +455,11 @@ func (track *VideoTrack) NewRTPReader(codecName string, ssrc uint32, mtu int) (R
 	}, nil
 }
 
+// returned encoderController might be nil
+func (track *VideoTrack) GetEncoderController() codec.EncoderController {
```
kevmo314, line 459: Can this be named `EncoderController()`? See https://google.github.io/styleguide/go/decisions.html#getters

Same for the other ones.

kevmo314 (commented):
Change seems reasonable, only a comment on the naming.

track.go
3DRX: Sure! I just pushed a commit to change the name. Thx for reviewing.

kevmo314 (approved):

PR #621, opened 2025-04-08
Add codec ffmpeg

3DRX:
This is a WIP PR for adding ffmpeg based video encoder.


.github/workflows/ci.yaml
```diff
@@ -52,34 +114,99 @@ jobs:
...
+          echo "=== FFmpeg build finished ==="
+          ls -la tmp/n7.0/
+          ls -la tmp/n7.0/lib/ 2>/dev/null || echo "lib directory not found after build"
+          ls -la tmp/n7.0/include/ 2>/dev/null || echo "include directory not found after build"
+      
+      - name: Verify FFmpeg installation
+        working-directory: pkg/codec/ffmpeg
+        run: |
+          ls -la tmp/n7.0/
+          ls -la tmp/n7.0/lib/ || echo "lib directory not found"
```
at-wat, line 169: `false || echo` will always success.
It should be like
```suggestion
          ls -la tmp/n7.0/lib/ || (echo "lib directory not found"; exit 1)
```

.github/workflows/ci.yaml
```diff
@@ -34,12 +36,72 @@ jobs:
...
+          pkg-config --exists --print-errors libavcodec || echo "pkg-config check failed"
+        env:
+          PKG_CONFIG_PATH: ${{ github.workspace }}/pkg/codec/ffmpeg/tmp/n7.0/lib/pkgconfig
+      
       - name: Run Test Suite
-        run: make test
+        run: |
+          make test \
+          && cd pkg/codec/ffmpeg \
+          && make test
```
at-wat, line 97: It might be better to separate the CI workflow for ffmpeg codec with path filter like
```yaml
name: CI
on:
  pull_request:
    branches:
      - master
    paths:
      - pkg/codec/ffmepg
  push:
    branches:
      - master
    paths:
      - pkg/codec/ffmepg
```

to make it easy to determine which of the main module and the ffmpeg codec module is failing.
Also, the main module build duration will not be affected by ffmpeg cache expiration by separating the workflow.

pkg/codec/ffmpeg/ffmpeg.go
```diff
@@ -0,0 +1,452 @@
...
+	frame          *astiav.Frame
+	packet         *astiav.Packet
+	width          int
+	height         int
+	r              video.Reader
+	nextIsKeyFrame bool
+
+	mu     sync.Mutex
+	closed bool
+}
```
at-wat, line 47: Maybe we can 
```suggestion
type baseEncoder struct {
	codecCtx       *astiav.CodecContext
	frame          *astiav.Frame
	packet         *astiav.Packet
	width          int
	height         int
	r              video.Reader
	nextIsKeyFrame bool

	mu     sync.Mutex
	closed bool
}

type hardwareEncoder struct {
	baseEncoder
  
	hwFramesCtx    *astiav.HardwareFramesContext
	hwFrame        *astiav.Frame
}

type softwareEncoder struct {
	baseEncoder

	codec          *astiav.Codec
}
```
to reuse the common members and functions between `hardwareEncoder` and `softwareEncoder`

pkg/codec/ffmpeg/ffmpeg.go
```diff
@@ -0,0 +1,452 @@
...
+
+	// Send frame to encoder
+	if err := e.codecCtx.SendFrame(e.hwFrame); err != nil {
+		return nil, func() {}, err
+	}
+
+	for {
+		if err = e.codecCtx.ReceivePacket(e.packet); err != nil {
+			if errors.Is(err, astiav.ErrEof) || errors.Is(err, astiav.ErrEagain) {
+				continue
```
at-wat, line 255: Will it be a busy loop or is there a sleep in `ReceivePacket`?

pkg/codec/ffmpeg/ffmpeg.go
```diff
@@ -0,0 +1,452 @@
...
+		e.nextIsKeyFrame = false
+	} else {
+		e.frame.SetPictureType(astiav.PictureType(astiav.PictureTypeNone))
+		e.hwFrame.SetPictureType(astiav.PictureType(astiav.PictureTypeNone))
+	}
+
+	err = e.frame.Data().FromImage(img)
+	if err != nil {
+		return nil, func() {}, err
+	}
```
at-wat, line 240:
```suggestion
	if err = e.frame.Data().FromImage(img); err != nil {
		return nil, func() {}, err
	}
```

pkg/codec/ffmpeg/ffmpeg_test.go
```diff
@@ -0,0 +1,357 @@
...
+			rel()
+			r.Controller().(codec.KeyFrameController).ForceKeyFrame()
+			_, rel, err = r.Read()
+			if err != nil {
+				t.Fatal(err)
+			}
+			// TODO: check if this is a key frame
+			// if !r.(*encoder).isKeyFrame {
+			// 	t.Fatal("Not a key frame")
+			// }
```
at-wat, line 219: Doesn't it work for now?

at-wat (commented):
Seems very nice!

3DRX:
@at-wat Thanks for reviewing! I will address them in seperate commits, for the easiness of review.

pkg/codec/ffmpeg/ffmpeg_test.go
3DRX: The commented out code is from tests of other encoders in this project, the problem here is that `isKeyFrame` is not a property in the newly implemented `softwareEncoder` and `hardwareEncoder`, we only have `nextIsKeyFrame` here so we can't just test it like this😅. This ffmpeg encoder do supports forcing key frame though. Perhaps we can parse the nalu header of the output data to verify if it's a key frame?

3DRX:
@at-wat I've resolved the conversations that I think may be resolved by follow up changes for less buzz. Feel free to reopen them if needed😊

3DRX:
@at-wat Sorry for the delay, I've tested and tweaked the codec config options and found that as long as the encoder is configured with constrained VBR and no B frames (suitable for realtime video), every `ReceivePacket` call should success. Therefore we can get rid of the for loop at all, and stop on any error returned.

Thank you so much for the review :)

PR #624, opened 2025-04-16
Add codeql-analysis.yml

3DRX:
#### Description

Don't know exactly why this is the case, but it seems like the missing of `codeql-analysis.yml` is causing warning in the code scanning secion.

image


JoeTurki:
I don't think this would work, we need a custom codeql config for this repo if we were to add one, also this wouldn't be easy because of the usage of native bindings, cgo, and C and C++ code.

There was an attempt to make one https://github.com/pion/mediadevices/pull/257

3DRX:
Oh, then I suppose this can be closed for now.

Issue #633, opened 2025-06-12
panic: send on closed channel


jackielii:
### Your environment.
  - Version: d9022069
  - Browser: N/A
  - Other Information - 

I built my own test track and close the read, and I get panic.

### What did you do?

I build a test track like this:

```
type pcmAudioSource struct {
	buf [][]byte
}

func (a *pcmAudioSource) Close() error { return nil }
func (a *pcmAudioSource) ID() string   { return "gemini-pcm-audio" }

// Read implements audio.Reader.
func (p *pcmAudioSource) Read() (chunk wave.Audio, release func(), err error) {
	if len(p.buf) == 0 {
		return nil, nil, io.EOF
	}
	data := p.buf[0]
	p.buf = p.buf[1:]

	time.Sleep(20 * time.Millisecond) // Simulate processing delay

	v := wave.NewInt16Interleaved(wave.ChunkInfo{
		Len:          len(data) / 2,
		Channels:     1,
		SamplingRate: 24000,
	})
	for i := range len(data) / 2 {
		sample := binary.LittleEndian.Uint16(data[i*2:])
		v.Data[i] = int16(sample)
	}
	return v, func() {}, err
}

type testAudioTrack struct{ buf chan []byte }

func newTestAudioTrack(ctx context.Context) mediadevices.Track {
	source := &pcmAudioSource{ctx: ctx}
	codecSelector := mediadevices.NewCodecSelector(mediadevices.WithAudioEncoders(&opus.Params{Latency: opus.Latency40ms}))
	track := mediadevices.NewAudioTrack(source, codecSelector)
        // read all samples and append into source.buf
        source.buf = append(source.buf, ...)
}
```

I then use this track in webrtc: `pc.AddTrack(newTestTrack())`

When I stop the peer connection in the browser, the panic occurs

### What did you expect?

No panic

### What happened?

I see error:

```sh
panic: send on closed channel

goroutine 43 [running]:
github.com/pion/mediadevices.(*baseTrack).unbind(0x140000c3570, {0x103698db0?, 0x1400043d490?})
	/Users/jackieli/personal/mediadevices/track.go:293 +0xa4
github.com/pion/mediadevices.(*AudioTrack).Unbind(0x14000338b28?, {0x103698db0?, 0x1400043d490?})
	/Users/jackieli/personal/mediadevices/track.go:515 +0x28
github.com/pion/webrtc/v4.(*RTPSender).ReplaceTrack(0x140000cc500, {0x0, 0x0})
[...]

PR #634, opened 2025-06-12
fix track.unbind panic

jackielii:
Fixes: #633

Here the signalCh could have been closed by another goroutine, we should
use returned signalCh from `track.removeActivePeerConnection()` to close
the channel.

Actually, I don't know why we need to close the signalCh, we're using it
to send over the doneCh, why ever close it?


3DRX (approved):
Thank you so much, this looks good. The failing ci seems to be irrelevant, let's merge this fix first and I'll try to fix ci later.

JoeTurki (approved):
lgtm

PR #635, opened 2025-06-13
Try to fix libvpx not found during ci test

3DRX:
Try to fix the following ci build error

https://github.com/pion/mediadevices/actions/runs/15627214233/job/44023636984

on macOS

```
ld: warning: search path '/opt/homebrew/Cellar/libvpx/1.15.1/lib' not found
ld: library 'vpx' not found
```

The libvpx installed is 1.15.2, but during build the compiler tries to find 1.15.1

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,6 +59,7 @@ jobs:
       - name: Install dependencies
         run: |
           which brew
+          brew update
           brew install \
             pkg-config \
             opus \


3DRX:
> `/opt/homebrew/Cellar/libvpx/1.15.1/lib` is determined via pkg-config. So, maybe brew installed libvpx has wrong `vpx.pc` file?

Good idea! I'll have a look, it's wierd that only go1.22 test is failing though.

3DRX:
@at-wat I purged go cache setup-go-macOS-arm64-go-1.22.12-89f1c539d83ecc5fa3278ce574acd1ab9e91b09c069abfde5afb6fd068bbfdb9 and re-run the job, and it successed. I don't know how go cache in github actions work exactly, but perhaps the cache hash don't include the version of c libraries used.

PR #648, opened 2025-09-04
[VPX] add vpx decoder

lkang-nuro:
#### Description
Add vpx decoder (VP8)


jiajieli-dev (approved):

pkg/codec/vpx/vpx_decoder.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (d *Decoder) Decode(data []byte) {
+	if len(data) == 0 {
+		return
+	}
+	status := C.decodeFrame(d.codec, (*C.uint8_t)(&data[0]), C.uint(len(data)))
+	if status != C.VPX_CODEC_OK {
+		fmt.Println("Decode failed", status)
+		panic("Decode failed")
```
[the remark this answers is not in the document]
lkang-nuro: Good point. Thanks for the catch. Added an error code.

pkg/codec/vpx/vpx_decoder.go
```diff
@@ -0,0 +1,111 @@
...
+import "C"
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/pion/mediadevices/pkg/prop"
+)
+
+type Decoder struct {
```
at-wat, line 60: All packages under `mediadevices/pkg/codec` implements the unified encoder interface which reads video/audio from `mediadevices/pkg/io/audio.Reader`(which reads `mediadevices/pkg/wave.Audio`)/`mediadevices/pkg/io/video.Reader`(which reads `image.Image`) and returns encoded bytes.
https://github.com/pion/mediadevices/blob/6047a32ea0fdbbf06ba0ea43296c7cccc7a5eff0/pkg/codec/codec.go#L125-L147

But this `Decoder` returns images in the codec specific format and the interface can't be shared among the codecs.
In my opinion, decoder interface should be defined at `mediadevices/pkg/codec` and the interface should be able to be shared among the codecs.

If this decoder really need to return the data in codec specific format, the name should be something like `RawDecoder` to leave `Decoder` for unified API available.

pkg/codec/vpx/vpx_decoder.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (d *Decoder) GetFrame() *VpxImage {
+	var iter C.vpx_codec_iter_t = nil // initialize to NULL to start iteration
+	img := C.getFrame(d.codec, &iter)
+	if img == nil {
+		return nil
+	}
+	return &VpxImage{img: img}
+}
```
at-wat, line 107: I assume the reason having `Decode/GetFrame` is that the number of the input byte stream and the output frames aren't always matched.
In this case, it's better to have a similar (symmetric) API with the encoder:
Something like
```
type VideoDecoderBuilder interface {
  BuildVideoDecoder(r io.Reader, p prop.Media) (video.Reader, error)
}
```

pkg/codec/vpx/vpx_image.go
```diff
@@ -0,0 +1,40 @@
+package vpx
+
+/*
+#cgo pkg-config: vpx
+#include <vpx/vpx_image.h>
+*/
+import "C"
+import "unsafe"
+
+type VpxImage struct {
```
at-wat, line 10: Users don't have a safe way to release the data.
Also to make the `Decoder` interface unified among the codecs, the decoder should return `image.Image`.

`mediadevices/io/video.Reader` returns release function along with `image.Image`.
So, the image implementing `image.Image` may have `C.vpx_image_t` internally and have release function written in CGO.


BTW, in Go's naming convention, it's `VPXImage`. https://go.dev/wiki/CodeReviewComments#initialisms

lkang-nuro:
> The interface looks good! Added a suggestion for handling multiple frames for one decode call without dropping.

Thanks @at-wat , all comments are super helpful and quite to the point. It seems all your comments are addressed, so I will go ahead and merge since we have other repo depends on this PR. Meanwhile, if you have any further comments, I am happy to open another PR to address. Thanks for understanding.

PR #652, opened 2025-10-02
Add VPX wrappers

hanguyen-nuro:
Needed for real-time MP4 generation in `bwe-test`. The `VpxImage` utility provides a clean Go interface to VPX image data, replacing complex unsafe pointer operations with simple methods like `Width()`, `Height()`, and `Plane()` for accessing decoded frame data needed in MP4 generation.

- Add vpx_image.go: VpxImage wrapper for vpx_image_t
- Move BitrateTracker to vpx package: More specific to VPX codec usage
- Add bitrate_tracker_test.go: Test coverage for VPX-specific bitrate tracking
- Remove generic codec-level BitrateTracker: Replaced by VPX-specific version

### DIFF
--- a/pkg/codec/vpx/bitrate_tracker.go
+++ b/pkg/codec/vpx/bitrate_tracker.go
@@ -1,0 +1,48 @@
+package vpx
+
+import (
+	"time"
+)
+
+type BitrateTracker struct {
+	windowSize time.Duration
+	buffer     []int
+	times      []time.Time
+}
+
+func NewBitrateTracker(windowSize time.Duration) *BitrateTracker {
+	return &BitrateTracker{
+		windowSize: windowSize,
+	}
+}
+
+func (bt *BitrateTracker) AddFrame(sizeBytes int, timestamp time.Time) {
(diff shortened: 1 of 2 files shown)


yzhao-nuro:
@hanguyen-nuro Does VpxImg structure hold raw image data such as YCbCr/RGBA? Could we use some native go Image library?

yzhao-nuro:
> > @hanguyen-nuro Does VpxImg structure hold raw image data such as YCbCr/RGBA? Could we use some native go Image library?
> 
> Thank you~ It holds raw image data in YUV format (specifically YCbCr) that one of the functions we need `ImageToMatNV12` is extracting this raw YUV data. Yep I think there are some alternative libraries that are more native. The current e2e working code in our internal repo implemented based on this function though.

Thanks! My thought is trying to use native go library when it is available instead of using CGO. Go Image library has image.YCbCr structure.

jiajieli-dev:
> Thanks! My thought is trying to use native go library when it is available instead of using CGO. Go Image library has image.YCbCr structure.

Make sense. @hanguyen-nuro we can timebox the investigation on using native go library.

PR #684, opened 2026-02-10
(3/3) Fix concurrency webcam windows

hexbabe:
## Summary

- **Per-thread COM initialization**: `CoInitializeEx` was only called once in `init()`. Go goroutines can run on any OS thread, so `Open()` and `VideoRecord()` now call `CoInitializeEx` on their current thread before making COM calls. `COINIT_MULTITHREADED` is idempotent so repeated calls are safe.
- **Mutex and idempotent Close**: Added `sync.Mutex` protecting shared camera state (`cam`, `closed`, `ch`, `done`) to prevent races between concurrent `Open`/`Close`/`VideoRecord`/`Properties` calls. `Close` snapshots fields under the lock, nils them out, then tears down outside the lock. Double close is a no-op.
- **Done channel for callback unblock**: `imageCallback` previously did a blocking send on `ch`. If no reader was consuming, this blocked the DirectShow streaming thread forever. Added a `done` channel to the select so `Close` can unblock it.
- **C-allocated frame buffer**: The DirectShow `BufferCB` callback writes into a frame buffer from a non-Go thread. Previously this buffer was a Go heap slice whose backing array the GC could relocate, causing silent memory corruption. Now allocated with `C.malloc` and freed explicitly in `Close`.
- **C++ bugfixes**: Fixed `delete cam->props` to `delete[] cam->props` (mismatched new[]/delete). Removed unused `LPOLESTR name` variable in `listResolution`.

Stacked on top of https://github.com/pion/mediadevices/pull/683

### DIFF
--- a/pkg/driver/camera/camera_windows.cpp
+++ b/pkg/driver/camera/camera_windows.cpp
@@ -5,6 +5,7 @@
 #include <dvdmedia.h>
 #include <qedit.h>
 #include <mmsystem.h>
+#include <wmcodecdsp.h>
 
 #include "camera_windows.hpp"
 #include "_cgo_export.h"
--- a/pkg/driver/camera/camera_windows.cpp
+++ b/pkg/driver/camera/camera_windows.cpp
@@ -52,6 +52,27 @@ void printErr(HRESULT hr)
 // returned pointer must be released by free() after use.
 char* getCameraName(IMoniker* moniker)
 {
+  // Try to get the friendly name via IPropertyBag.
+  IPropertyBag* propBag = nullptr;
+  if (SUCCEEDED(moniker->BindToStorage(0, 0, IID_IPropertyBag, (void**)&propBag)))
+  {
+    VARIANT varName;
+    VariantInit(&varName);
+    if (SUCCEEDED(propBag->Read(L"FriendlyName", &varName, 0)))
+    {
+      std::string nameStr = utf16Decode(varName.bstrVal);
+      VariantClear(&varName);
+      propBag->Release();
+
+      char* ret = (char*)malloc(nameStr.size() + 1);
+      memcpy(ret, nameStr.c_str(), nameStr.size() + 1);
+      return ret;
+    }
+    VariantClear(&varName);
+    propBag->Release();
+  }
+
+  // Fallback to display name.
   LPOLESTR name;
   if (FAILED(moniker->GetDisplayName(nullptr, nullptr, &name)))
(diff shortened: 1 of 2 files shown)


hexbabe:
Hi @JoTurk this will be among my final contributions to mediadevices for a while :,)

could you take a look this week?

JoTurk:
> this will be among my final contributions to mediadevices for a while

:(

thank you for all the work you've done.

> could you take a look this week?

Yeah, I'll try it to test it today or tomorrow, I have a 2nd machine with Windows, So testing and reviewing these PRs is a lot easier for me now :)

pkg/driver/camera/camera_windows.cpp
```diff
@@ -253,7 +253,6 @@ int listResolution(camera* cam, const char** errstr)
   ICaptureGraphBuilder2* captureGraph = nullptr;
   IAMStreamConfig* config = nullptr;
   IPin* src = nullptr;
-  LPOLESTR name;
```
hexbabe, line 256: unused, not needed

pkg/driver/camera/camera_windows.go
```diff
@@ -106,31 +126,68 @@ func imageCallback(cam uintptr) {
...
+	if done != nil {
+		close(done)
+	}
+
+	if cam != nil {
+		C.free(unsafe.Pointer(cam.name))
+		C.freeCamera(cam)
+	}
+
+	C.free(cbuf)
```
JoTurk, line 166: Wouldn't this run into a potential UAF if we have an already running imageCallback callback?

hexbabe: This is actually safe because `freeCamera` calls `IMediaControl::Stop()`, which blocks until any in-flight `BufferCB` returns. `close(done)` before freeCamera ensures the callback can't deadlock in the select. So `C.free(cbuf)` only runs after all callbacks are done.

That said, I moved the `RUnlock` past the copy so the safety is obvious from the Go side without needing to know DirectShow's `Stop()` semantics.

hexbabe: https://github.com/pion/mediadevices/blob/63c09cb1f45bc7c40b99022bda35dd720294229d/pkg/driver/camera/camera_windows.cpp#L612

hexbabe:
Hey @JoTurk  could I get a second pass review here? TY

JoTurk (approved):
Thank you

PR #712, opened 2026-08-27
codec/opus: fix potential panic on encode failure

GerardGao:
`encoder.Read` calls `opus_encode`/`opus_encode_float` and then unconditionally evaluates `encoded[:n:n]`. When the encode fails, libopus returns a negative error code (e.g. `OPUS_BAD_ARG`, `OPUS_BUFFER_TOO_SMALL`), so `n` is negative and the slice expression panics with `slice bounds out of range [::-1]` instead of returning an error to the caller.

## Change

- Return an error (including the libopus error code) when `n < 0`, instead of slicing with a negative index.
- Add `-tags opustest` to the `go vet` and `go test` targets in the Makefile so the regression test below actually runs in CI.

## Tests

- New `TestReadEncodeErrorDoesNotPanic` (build tag `opustest`) feeds an invalid frame size (121 samples at 48kHz; must be a multiple of 120) which deterministically makes `opus_encode` return `OPUS_BAD_ARG`.
  - Before the fix: `panic: runtime error: slice bounds out of range [::-1]` at `opus.go:119`.
  - After the fix: returns an error, no panic.
- `go test -v -race -tags opustest ./pkg/codec/opus/` passes, including the pre-existing encoder tests.
- `go test -race ./pkg/codec/opus/` (no tag) passes, confirming the build-tag isolation does not affect normal builds.


pkg/codec/opus/read_error_test.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// TestReadEncodeErrorDoesNotPanic reproduces a panic in encoder.Read: when
+// opus_encode fails (returns a negative error code), the code still evaluates
+// encoded[:n:n] with a negative index, which panics with "slice bounds out of
+// range". The expected behavior is to return the error instead.
+//
+// Trigger: opus_encode returns OPUS_BAD_ARG when frame_size is invalid. At
+// 48kHz the minimum frame size is 120 samples and it must be a multiple of
+// 120, so 121 samples deterministically produces a negative return value.
```
at-wat: As this package always wraps the reader with `audio.Buffer`, source readers with any frame size never trigger such a panic in practice.
This test doesn't look like it's reproducing a real problem; it's just implementing a new buggy alternative and testing it.

I do agree with the change in `opus.go`, but the test doesn't look proper.
I would recommend updating only `opus.go` to fix a potential bug that doesn’t have actual reproduction for now.

at-wat (commented):
Thanks for your contribution

pkg/codec/opus/read_error_test.go
at-wat: If there is an actual case causing a panic, it would be nice to reproduce it with a test.

GerardGao: You're right. I verified the two ways opus_encode can return a negative value, both through the public API: 1) OPUS_BAD_ARG (invalid frame size): audio.Buffer outputs Latency.samples(SampleRate) samples, and every validated latency is a multiple of opus's frame-size requirements, so this can't happen. 2) OPUS_BUFFER_TOO_SMALL: with the default 20ms latency, 510kbps bitrate and white noise (RFC 6716 allows up to 1275 bytes for a 20ms frame), the largest output over 200 frames was exactly 1024 bytes - libopus adapts to the max_data_bytes argument, so this can't happen either. The test has been removed; the PR now only keeps the defensive change in opus.go.

GerardGao: I checked both negative-return paths through the public API and neither is reachable today: audio.Buffer guarantees a valid frame size (OPUS_BAD_ARG impossible), and the 1024-byte max_data_bytes argument makes libopus adapt its output (OPUS_BUFFER_TOO_SMALL impossible - measured 200 frames at 510kbps with white noise, max output exactly 1024 bytes). The panic would only resurface if those invariants changed, so the defensive fix in opus.go is kept without a test.

at-wat (approved):
LGTM

PR #713, opened 2026-08-28
vpx: fix codec context leak on resolution change

GerardGao:
When a resolution change is detected, `encoder.Read` reinitializes the encoder context: it allocates a new `vpx_codec_ctx_t` shell, runs `vpx_codec_enc_init_ver`, and then releases the old context with plain `free()`.

`vpx_codec_enc_init` allocates internal state (lookahead buffers, reference frame buffers, etc.) that only `vpx_codec_destroy` releases. Freeing just the shell therefore leaks the entire internal state of one encoder context on every resolution change.

Measured with a heap-usage test: 40 resolution switches leak 31.6 MB (~790 KB per switch) before the fix, and 0 bytes after.

The pattern was introduced in #283 ("Fix VPX_CODEC_INVALID_PARAM on resolution change"); `Close()` already uses the correct destroy-then-free order, so this was an oversight, not a deliberate choice.

## Change

- Call `vpx_codec_destroy` on the old context before `free()` in the resolution-change path, matching the order in `Close()`.
- Free the `newCtx` shell on the `enc_init` failure path (it was leaked there too).
- Add `-tags vpxtest` to the `go vet` and `go test` targets in the Makefile so the regression test runs in CI.

## Tests

- New `TestResolutionChangeDoesNotLeakOldCodecContext` (build tag `vpxtest`) alternates 16x16 and 32x32 frames, which forces the reinit path, and measures live heap bytes around the switches:
  - Before the fix: FAIL, 31.6 MB leaked across 40 switches.
  - After the fix: PASS, 0 bytes leaked.
- `go test -race -tags vpxtest ./pkg/codec/vpx/` passes, including the existing `TestEncoder` and `TestImageSizeChange` suites.
- `go test ./pkg/codec/vpx/` (no tag) passes, confirming build-tag isolation does not affect normal builds.

Note: the test measures the process malloc heap because libvpx 1.15+ removed the `vpx_mem_set_functions` injection hook and allocates directly from the system allocator. It uses `malloc_zone_statistics` on Darwin and `mallinfo2` on Linux.


pkg/codec/vpx/vpxtest_helper.go
```diff
@@ -0,0 +1,33 @@
+//go:build vpxtest
```
at-wat: Adding tags for each test with cgo makes package maintenance harder.
Adding an internal package like `github.com/pion/mediadevices/internal/cgoheap` and importing it from `_test.go` would be better.

I think this should be OS conditional like `//go:build linux`.

pkg/codec/vpx/resolution_leak_test.go
```diff
@@ -0,0 +1,99 @@
...
+	img := image.NewYCbCr(image.Rect(0, 0, w, h), image.YCbCrSubsampleRatio420)
+	for i := range img.Y {
+		img.Y[i] = 128
+	}
+	for i := range img.Cb {
+		img.Cb[i] = 128
+	}
+	for i := range img.Cr {
+		img.Cr[i] = 128
+	}
```
at-wat: Contents of the image don't matter. Initial all 0 would be fine.

pkg/codec/vpx/codec_context_leak_test.go
at-wat, line 1: Better to be like `codec_context_leak_test.go`

pkg/codec/vpx/vpx.go
```diff
@@ -283,8 +283,17 @@ func (e *encoder) Read() ([]byte, func(), error) {
 		if ec := C.vpx_codec_enc_init_ver(
 			newCodec, e.codec.iface, e.cfg, 0, C.VPX_ENCODER_ABI_VERSION,
 		); ec != 0 {
-			return nil, func() {}, fmt.Errorf("vpx_codec_enc_init failed (%d): %s", ec, C.GoString(C.error_detail_safe(newCodec)))
+			// Capture the error detail before freeing the shell allocated by
+			// newCtx. vpx_codec_destroy is not needed here because enc_init
+			// failed and did not acquire internal resources.
```
at-wat: Looks like it's explaining the change, not the final code.
Please keep the comment explaining only the current code, and leave the comment for the change as a PR comment or self-review comment.

pkg/codec/vpx/vpxtest_helper.go
GerardGao: Done. Heap measurement now lives in internal/cgoheap (darwin/linux implement it, other platforms return 0, and a nocgo stub keeps CGO_ENABLED=0 builds working). The vpx test imports it from _test.go with no build tag, and the Makefile change is reverted.

pkg/codec/vpx/resolution_leak_test.go
GerardGao: Done. Frames are now image.NewYCbCr with zero content.

pkg/codec/vpx/codec_context_leak_test.go
GerardGao: Done. Renamed to codec_context_leak_test.go.

pkg/codec/vpx/vpx.go
GerardGao: Done. The comments now describe only what the current code does: the failed init did not acquire internal resources so the shell can be freed directly, and the old context must be destroyed to release its internal state.

fallenmi (changes requested):
`internal/cgoheap/heap_linux.go` is selected for every Linux+cgo build, but it unconditionally uses `struct mallinfo2` / `mallinfo2()`. `mallinfo2` was introduced in glibc 2.33 and is absent from musl's `<malloc.h>`; it is not a generic Linux API. Because the new untagged VPX test imports this package, the entire `pkg/codec/vpx` test binary no longer compiles on Alpine/musl.

I reproduced this on exact head `72f87f4da5c281ac17542e95b2f3b4847c975c86` in an official Alpine 3.23.4 aarch64 VM with musl 1.2.5, Go 1.25.10, and libvpx 1.15.2. `go test -v -run '^TestImageSizeChange$' ./pkg/codec/vpx` passes on exact base `29a80cc3a982a22e1acc283aa15339638f9ff767` (all VP8/VP9 resize cases), but exact head and GitHub synthetic merge `0638569c24c255e68a08abb4ec674b04d7bd829f` fail at `heap_linux.go:12` with incomplete `struct mallinfo2` and an implicit `mallinfo2` declaration. The same unconditional API use also excludes glibc before 2.33.

Please guard the allocator probe for the actual libc/API version and provide a portable fallback (or keep this measurement behind an explicit test build tag), and add a musl/old-glibc compile regression.

The underlying production fix is sound: my independent exact base/head leak oracle measured +31,587,200 bytes on base in 3/3 runs and passed head 3/3, while the submitted leak test passed 5/5 with zero growth. The exact synthetic merge also passes the full macOS `make test` path, including vet, CGO/no-CGO builds, example builds, and race-enabled tests; it is specifically the broad Linux libc assumption that blocks this head.

AI disclosure: I used OpenAI Codex to inspect the libvpx lifecycle and platform build paths, construct and run the exact base/head leak oracle, reproduce the clean-base/head/synthetic-merge build on Alpine/musl, execute the full macOS test suite, and draft this review. I manually verified the code paths, compiler diagnostics, exact commits, public policy/interactions, and conclusion.

GerardGao:
Thanks for the detailed review. The musl/old-glibc build break is fixed in 96311d7:

1. `internal/cgoheap/heap_linux.go` now guards `mallinfo2` with `__GLIBC_PREREQ(2, 33)` and falls back to `mallinfo` elsewhere. On musl `mallinfo` is a stub whose `uordblks` is 0, so the leak test degrades to a no-op assertion there (it still passes) while glibc >= 2.33 keeps the exact measurement.
2. Added a `build-alpine` CI job (alpine:3.23 container, musl + libvpx-dev) that runs `go build ./internal/cgoheap/` and the vpx leak/resize tests on musl, covering the regression you reproduced.

Note: I couldn't run an Alpine container locally (no Docker here), so the new CI job is the primary verification for the musl path — flagging that explicitly in case the job config needs adjustment for the Actions runner.

GerardGao:
Follow-up: I installed Docker locally and reproduced your Alpine failure, then verified the fix in containers.

**Reproduction**: my previous fallback was wrong too — musl's malloc.h provides neither `mallinfo2` nor `mallinfo`. Confirmed in alpine:3.23 (aarch64, Go 1.25.10, libvpx 1.15.2, same as your setup): compile failed at the `struct mallinfo` declaration.

**Fix (a7d2c72)**:
- `mallinfo2` is now gated on `__GLIBC_PREREQ(2, 33)`, `mallinfo` on `__GLIBC__`, and the probe returns 0 on other libcs (musl).
- Added `cgoheap.Supported()`; the leak test now skips when the platform lacks heap accounting instead of asserting against zeros.

**Verified in local containers**:
1. alpine:3.23 (musl 1.2.5, aarch64): `go build ./internal/cgoheap/` passes; `TestImageSizeChange` passes all 6 sub-cases; the leak test skips cleanly.
2. debian:bookworm (glibc 2.36): the `mallinfo2` path executes for real — leak test runs (not skipped) and passes with ~4.7KB/switch.
3. macOS host: leak test runs with 0 bytes/switch; `go test -race ./pkg/codec/vpx/`, vet, gofmt, and the CGO_ENABLED=0/windows cross-builds all pass.

The Alpine CI job added earlier stays as the standing musl regression.

GerardGao:
@fallenmi Thanks again for the thorough review. All four points are addressed: heap measurement moved to `internal/cgoheap`, frames use zero-content `image.NewYCbCr`, the test file was renamed, and the `vpx.go` comments now describe only current code. The only CI failure was an unrelated `svtav1` SIGABRT on Linux Go 1.25, which passed on re-run; all checks are green now. Could you take another look when you have a moment?

fallenmi (approved):
The previous portability blocker is resolved on exact head `4c60d092ab8bc195f3f7627e7cd87185108283d6`.

I rechecked the libc guard with a three-case oracle: it reports unsupported when neither API exists (musl), selects `mallinfo` for a glibc 2.32-like header, and selects `mallinfo2` for glibc 2.36; the old unguarded control still fails against the musl-like header. On the exact head and current merge `975068fd5bc9a6a6d56d0ed9206828c3524ed38e`, the submitted leak and resize tests passed 3/3 with zero measured live-heap growth.

The new Alpine job checked out that same merge: the cgoheap package built, the allocator-dependent leak test skipped as intended on musl, and all six VP8/VP9 resize cases passed. The independently recomputed merge tree matches GitHub, and all 7 checks across 2 suites / 1 workflow are green. Approved.

Disclosure: I used OpenAI Codex to reconstruct the prior finding and author response, inspect the libc guards and test paths, run the focused exact-head/current-merge checks and independent libc oracle, and verify the current merge, policy, interactions, and CI. I manually verified the exact revisions, evidence, and conclusion before submission.

.github/workflows/ci.yaml
```diff
@@ -80,3 +80,17 @@ jobs:
...
+    steps:
+      - name: Install dependencies
+        run: |
+          apk add --no-cache go gcc musl-dev libvpx-dev pkgconf make git nodejs
+      - name: Checkout
+        uses: actions/checkout@v7
+      - name: Build and test vpx on musl
+        run: |
+          go build ./internal/cgoheap/
+          go test -v -run 'TestImageSizeChange|TestResolutionChangeDoesNotLeakOldCodecContext' ./pkg/codec/vpx/
```
at-wat: This tests a regression introduced and fixed in this PR.
It will be unclear and less meaningful once this is merged.
If you want to add a CI job to test on Alpine, it should test the whole package.

(Similar to https://github.com/pion/mediadevices/pull/713#discussion_r3878376253, LLMs often try to add self-justification for the changes, which doesn't make sense after merging and increases noise in the code. Please provide proper instructions to the LLM, and always check everything yourself if you use it.)

.github/workflows/ci.yaml
```diff
@@ -80,3 +80,17 @@ jobs:
...
       - name: Checking licenses
         run: go-licenses check ./...
+  build-alpine:
+    runs-on: ubuntu-latest
+    container: alpine:3.23
+    name: Alpine musl (cgoheap + vpx regression)
+    steps:
+      - name: Install dependencies
+        run: |
+          apk add --no-cache go gcc musl-dev libvpx-dev pkgconf make git nodejs
```
at-wat: A version specification that is not consistent with other jobs is being added.
Go version should be controlled in the same manner as the other jobs if this job should be added.
e.g.:
```
    strategy:
      fail-fast: false
      matrix:
        go: ["1.26", "1.25"] # auto-update/supported-go-version-list
    container: golang:${{ matrix.go }}-alpine
```

GerardGao:
Thanks for the review, @at-wat!

Both issues addressed in 9547368:

1. **CI job scope**: Removed the `-run` filter — the Alpine job now runs `go test -v ./pkg/codec/vpx/` (entire package) instead of just the two regression tests.
2. **Go version management**: Replaced the fixed `alpine:3.23` container with a matrix over `["1.26", "1.25"]` using `golang:${{ matrix.go }}-alpine`, consistent with the other CI jobs. Dropped `go` from `apk add` since the `golang:*` image ships with it.

.github/workflows/ci.yaml
```diff
@@ -80,3 +80,21 @@ jobs:
...
+    name: Alpine musl Go ${{ matrix.go }}
+    steps:
+      - name: Install dependencies
+        run: |
+          apk add --no-cache gcc musl-dev libvpx-dev pkgconf make git nodejs
+      - name: Checkout
+        uses: actions/checkout@v7
+      - name: Build and test vpx on musl
+        run: |
+          go build ./internal/cgoheap/
```
at-wat, line 99: This line is not needed

.github/workflows/ci.yaml
```diff
@@ -80,3 +80,21 @@ jobs:
...
+          go test -v ./pkg/codec/vpx/
```
at-wat, line 100: The whole mediadevices package should be tested, not just the vpx sub-package, if this test job is added.
The regression in the vpx sub-package test has been introduced and fixed in this PR, and no clear reason will be visible once this PR is merged. So, there is no clear reason to test only the vpx sub-package on the main branch.
(similar to https://github.com/pion/mediadevices/pull/713#discussion_r3878376253)

.github/workflows/ci.yaml
```diff
@@ -80,3 +80,21 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        go: ["1.26", "1.25"] # auto-update/supported-go-version-list
+    container: golang:${{ matrix.go }}-alpine
+    name: Alpine musl Go ${{ matrix.go }}
+    steps:
+      - name: Install dependencies
+        run: |
+          apk add --no-cache gcc musl-dev libvpx-dev pkgconf make git nodejs
```
at-wat, line 94: The runner host provides the node runtime by volume mount. No need to install nodejs.
See https://github.com/pion/mediadevices/actions/runs/33708368367/job/100502352890#step:2:41

```suggestion
        run: apk add --no-cache gcc musl-dev libvpx-dev pkgconf make git
```

GerardGao:
Thanks @at-wat — all three points are addressed, and the branch is rebased onto latest `main`.

- Dropped `nodejs`: the runner mounts the node runtime, as your linked job log shows.
- Dropped `go build ./internal/cgoheap/`: `make test` already builds every package.
- The job now runs `make test`, the same entry point as `build-linux` and `build-darwin`, so the whole module is tested on musl rather than only `pkg/codec/vpx`. The step is renamed to `Run Test Suite` to match the other jobs, and the dependency list now mirrors what `build-linux` installs (`opus-dev`, `x264-dev`, `svt-av1-dev`, `libva-dev`, `libx11-dev`, `libxext-dev`) plus `g++`, which the openh264 binding needs.

I also dropped the `chore: retrigger CI` commit. The five vpx/cgoheap commits are unchanged.

To be explicit about what this job does and does not cover: `cgoheap.Supported()` is false on musl, because musl provides neither `mallinfo2` nor `mallinfo`. So `TestResolutionChangeDoesNotLeakOldCodecContext` is **skipped** on Alpine, not passing. The job's value is portability rather than leak detection — the `#include <malloc.h>` / `mallinfo2` code added in this PR does not compile on musl unless it is guarded, which was the blocker raised earlier in this review, and a glibc-only CI matrix cannot catch a regression there. Everything else in the module runs normally.

Verified locally in both matrix containers using the exact dependency list from the workflow:

- `golang:1.25-alpine`: `make test` exit 0
- `golang:1.26-alpine`: `make test` exit 0 — 103 tests pass, 0 failures, `pkg/codec/vpx` at 89.8% coverage, `TestImageSizeChange` passing, `TestResolutionChangeDoesNotLeakOldCodecContext` skipped as described above

That said, I agree this CI job is a separate concern from the vpx fix. If you would prefer to keep this PR strictly to the codec context leak, I am happy to drop the Alpine job here and open it as its own PR — just let me know which you prefer.

at-wat:
> That said, I agree this CI job is a separate concern from the vpx fix. If you would prefer to keep this PR strictly to the codec context leak, I am happy to drop the Alpine job here and open it as its own PR — just let me know which you prefer.

@GerardGao I think it's better to drop the Alpine job in this PR.
This was introduced to verify failure in the test. So it tests the test not the product.

